PR #4, opened 2016-01-24
DB Schema

andela-oojewale:
Design DB UML.
  Generate models and migration.
  Add relationships.
  NB: DB contains two polymorphic tables; comments and tags.


andela-ooranagwa:
:+1:

andela-aadepoju:
:+1:

db/schema.rb
```diff
@@ -0,0 +1,64 @@
...
+    t.string   "subscriber_type"
+    t.datetime "created_at",      null: false
+    t.datetime "updated_at",      null: false
+  end
+
+  add_index "tags", ["subscriber_type", "subscriber_id"], name: "index_tags_on_subscriber_type_and_subscriber_id"
+
+  create_table "users", force: :cascade do |t|
+    t.string   "uuid"
+    t.string   "provider"
```
andela-oakinniranye: Are we now hard-coding the provider and uuid to the User object?

db/migrate/20160124145702_create_comments.rb
```diff
@@ -0,0 +1,12 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.integer :user_id
+      t.string :content
+      t.integer :votes
+      t.references :comment_on, polymorphic: true, index: true
```
andela-oakinniranye: Not much of a comment, but a convention thingy! Why using comment_on? Thought it would be clearer for anyone taking on the project to see commentable(has_many comments as: :commentable and belongs_to  :commentable)

andela-oojewale: Good observation. I just felt in normal grammar. "This is a `comment_on` answer or question" seems more readable than "This comment is `commentable` on answer or question".

What do you think. Should we still go with `commentable` or leave it as it is?

db/schema.rb
andela-oojewale: `hard-coding`? I don't get exactly what that means in this context. Could you please clarify?

Also, I thought we postponed issues with user information amendment and authentication till sometime later.

andela-oakinniranye: Sorry, I could have clarified that better!
I meant are we making the uuid and provider static?
Because if that were the case, for each sign in with a provider it would be creating a new user.
Yes, we agreed that we were going to postpone issues with user info etc. If we place those columns there now, it may become more of a legacy thing `(where no one knows why it was placed and doesn't want to touch, but its not really serving any purpose)`. What I'm saying is if its not useful now, we can remove it till it is.

andela-ooranagwa: My take (unless I'm missing something, which isn't unlikely) is that since, for now, we are focusing on one provider; `Google` we might not have need for the provider. However, the `uuid` is necessary if we are to keep track of user resources (like `comments`, `questions`, etc).

andela-oojewale: Fine, I get what you mean now. What column would be the relation column between this and the `possible` user_provider table? So, I could just add it, drops these columns and re-push.

PR #5, opened 2016-01-24
Create API routes using endpoints docs. Update endpoints docs to show…

andela-ooranagwa:
… available routes


andela-oakinniranye:
:+1: But you may want to remove the commented routes in `routes.rb` file though!

PR #7, opened 2016-01-28
Add TokenManager for generating and authenticating requests using JWT…

andela-ooranagwa:
… tokens


andela-oakinniranye:
@andela-oojewale There is no performance benefit far as I know, just that one throws an error, which if not handled well, may break your app, while the other just returns `nil`. Always better to handle `nil` than exception

andela-oakinniranye:
@andela-ooranagwa Should we bother about a user signing in from multiple clients(web app, mobile app, or multiple web apps etc) now? if so, we may need to move the `active` away from the users table, probably to a token table where all tokens issued to a user may be disabled etc. Else it is ok as is. and i think it should be good to merge

andela-oakinniranye:
:+1:

andela-oojewale:
:+1:

andela-oakinniranye:
Probably create a **services** or **managers** folder and add the `token_manager` there. I don't think it's supposed to be in the `controllers`

PR #8, opened 2016-01-28
Ft comment resources

andela-oojewale:


andela-oakinniranye:
Nice one @andela-oojewale! I think this is ok, but you may want to work on handling the chained methods e.g `resource.find_by(id: resource_id).comments` if the resource is not found, you would be chaining `comments` to `nil`, which would still return the exception you tried to escape. It's ok, but you may need to refactor after the merge :+1:

andela-oojewale:
@andela-ooranagwa, that is handled with the rescue block in the method body. Moreover, it's been changed to `find_by` for sometime. You might want to `compare` the branch with master if you don't the latest state is not revealed in the PR thread you're seeing.

andela-oakinniranye:
:+1: Saw these yesterday

andela-oakinniranye:
:+1:

andela-ooranagwa:
:+1:

PR #11, opened 2016-01-29
Ft votes resources

andela-oojewale:


app/controllers/votes_controller.rb
```diff
@@ -0,0 +1,32 @@
+class VotesController < ApplicationController
```
andela-ooranagwa: You're creating a unique style for yourself using `attr_reader` and `custom_initializer`

app/models/vote.rb
```diff
@@ -0,0 +1,27 @@
...
+          subject_of_vote.votes.create(user_id: user_id)
+          return subject.find_by(id: subject_id).votes.count
+        end
+      end
+      nil
+    end
+
+    def voted?(subject, subject_id, user_id)
+      vote = find_by(user_id: user_id, voteable_type: "#{subject}", voteable_id: subject_id)
+      vote.nil? ? false : true
```
andela-ooranagwa: Any need for explicitly returning `true` or `false`?

I remembered this, instead of fetching a whole object to check for existence, we could use `:exists?` 
> Vote.exists?(user_id: 1, voteable_type: "type", voteable_id: 1) #=> true or false

spec/controllers/comments_controller_spec.rb
```diff
@@ -51,4 +53,5 @@
     end
   end
 
+>>>>>>> master
```
andela-ooranagwa: lolz! :sweat_smile:

app/controllers/votes_controller.rb
andela-oojewale: I get the literal meaning of the statement but I don't get what you mean.

app/models/vote.rb
andela-oojewale: This is a great idea. Its been implemented. Thanks.

spec/controllers/comments_controller_spec.rb
andela-oojewale: Lol.

andela-ooranagwa:
:+1:

andela-oakinniranye:
:+1:

andela-aadepoju:
:+1:

PR #12, opened 2016-01-29
User implementation

andela-oakinniranye:


db/schema.rb
```diff
@@ -11,12 +11,13 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160128233820) do
+ActiveRecord::Schema.define(version: 20160129095819) do
 
   create_table "answers", force: :cascade do |t|
     t.integer  "user_id"
     t.integer  "question_id"
     t.string   "content"
+    t.integer  "votes"
```
andela-oojewale: Thought we would not need this again when we created the votes model. Is there anything I'm missing or have missed?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    # add meta data for pagination
+    # total: resource.total_entries,
+    # total_pages: resource.total_pages,
+    # first_page: resource.current_page == 1,
+    # last_page: resource.next_page.blank?,
+    # previous_page: resource.previous_page,
+    # next_page: resource.next_page,
+    # out_of_bounds: resource.out_of_bounds?,
+    # offset: resource.offset
```
andela-oojewale: Someone has forgotten to remove their comments. :grinning:

test/controllers/answers_controller_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test_helper'
+
+class AnswersControllerTest < ActionController::TestCase
```
andela-oojewale: If all these tests have passed, I don't need to write mine again. But I forgot to mention it the last time, I changed our test framework to rspec. Right from the beginning when I pulled. @andela-ooranagwa  and @andela-aadepoju are aware though.

app/models/question.rb
```diff
@@ -3,11 +3,15 @@ class Question < ActiveRecord::Base
   has_many :tags, as: :subscriber
   has_many :answers
   belongs_to :user
-
   validates :content, presence: true
 
   include Modify
 
+  def self.with_answers
```
andela-oojewale: I think this could be in the `class << self` scope below, since it's a class method as well.

config/application.rb
```diff
@@ -22,5 +22,8 @@ class Application < Rails::Application
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
     config.active_record.raise_in_transactional_callbacks = true
+
+    config.middleware.insert_after(ActiveRecord::QueryCache, ActionDispatch::Cookies)
+    config.middleware.insert_after(ActionDispatch::Cookies, ActionDispatch::Session::CookieStore)
```
andela-oojewale: Is this because there was need for session @ some point?

db/schema.rb
andela-oakinniranye: sure thing, That's probably from the last merge: I'm sure we've not added a migration to remove the votes from either answers/questions/comments.

app/controllers/users_controller.rb
andela-oakinniranye: This is not a comment, it's going to be the metadata that would be useful for the paginations, from `will_paginate`. Thought it would be useful to anyone that may want to add it to their serializer

test/controllers/answers_controller_test.rb
andela-oakinniranye: ok, that's rails default from `rails g controller` though

app/models/question.rb
andela-oakinniranye: sure thing

config/application.rb
andela-oakinniranye: Yeah. For authentication. Omniauth doesn't work without sessions

db/schema.rb
andela-oojewale: I added the migration but can't remember if it is on `vote` or `comment` branch.

app/controllers/users_controller.rb
andela-oojewale: Oh! I see.

config/application.rb
andela-oojewale: True.

andela-oojewale:
:+1:

app/controllers/answers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def show
+    render json: @answer
+  end
+
+  def create
+    @answer = @question.answers.new(answer_params)
+
+    if @answer.save
+      render json: @answer, status: :created, location: question_answer_path(@question, @answer)
```
andela-ooranagwa: Nice! Keeping to standards :+1:

app/controllers/answers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    else
+      render json: @answer.errors, status: :unprocessable_entity
+    end
+  end
+
+  def update
+    if @answer.update(answer_params)
+      head :no_content
+    else
+      render json: @answer.errors, status: :unprocessable_entity
```
andela-ooranagwa: @andela-oojewale raised concerns about returning form errors explicity.  Is there a better a way of doing this?

app/controllers/answers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  private
+
+    def set_answer
+      @answer = @question.answers.find(params[:id])
+    end
+
+    def set_question
+      @question = Question.with_answers.find(params[:question_id])
```
andela-ooranagwa: L45 & L41 is `find` which might fail if the give `id` doesn't exist.

app/controllers/answers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def destroy
+    @answer.destroy
+    head :no_content
+  end
+
+  private
+
+    def set_answer
+      @answer = @question.answers.find(params[:id])
```
andela-ooranagwa: For questions with _empty_ answers, _calling_ `find` on the result might result in an error.

app/controllers/application_controller.rb
```diff
@@ -1,14 +1,43 @@
 class ApplicationController < ActionController::API
-  def current_user
-    @current_user || authenticate_token
+  include ActionController::HttpAuthentication::Token::ControllerMethods
+
+  attr_reader :current_user
+  before_action :authenticate
```
andela-ooranagwa: Using the `:before_action :authenticate` method here; would imply that controllers that expose _public_ endpoints should have a `:skip_before_action` e.g. the `UsersController`

app/controllers/concerns/provider_helper.rb
```diff
@@ -0,0 +1,13 @@
+module ProviderHelper
+  def get_provider(provider)
+    provider_url = case provider
+    when /^google/
+      '/auth/google_oauth2'
+    when /^slack/
+      '/auth/slack'
+    else
+      raise "Invalid provider"
```
andela-ooranagwa: I seem not to find where this error is being caught. Could you help?

app/controllers/application_controller.rb
andela-oakinniranye: yes. That way we'd most certain of exposing routes that should be public

app/controllers/users_controller.rb
andela-ooranagwa: Yea, I agree but you don't seem to have used

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  def show
+    render json: @user, status: :ok
+  end
+
+
+  def update
+    if @user.update(user_params)
+      head :no_content
+    else @user.errors
+      render json: @user.errors, status: :unprocessable_entity
```
andela-ooranagwa: See above comment.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  def destroy
+    @user.destroy
+    head :no_content
+  end
+
+  def login
+    redirect_url = params[:redirect_url] || params[:rdr] || params[:redir]
```
andela-ooranagwa: Any reason for the multiple ways for providing the `redirect url`

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    provider = get_provider(params[:provider] || params[:p])
+    set_attrs_in_session({redirect_url: redirect_url})
+
+    redirect_to provider
+  end
+
+  def authenticate
```
andela-ooranagwa: Method name already exists in `Application controller` albeit `private` #just-saying

app/models/user.rb
```diff
@@ -3,8 +3,25 @@ class User < ActiveRecord::Base
...
+    user = where(email: auth.info.email).first_or_create do |u|
+      u.name= auth.info.name
+      u.email= auth.info.email
+    end
+    user.social_providers.from_social(auth)
+    user
+  end
+
+  def refresh_token
+    token = TokenManager.generate_token(self.id)
```
andela-ooranagwa: Any need storing it in a variable?

config/routes.rb
```diff
@@ -13,15 +14,21 @@
...
+  get 'users/renew_token' => 'users#renew_token'
   get "tags" => "tags#index"
   get "tags/popular" => "tags#popular"
   get "tags/recent" => "tags#recent"
   get "tags/trending" => "tags#trending"
+  
+  get 'login', to: "users#login"
+  get 'login/:provider', to: "users#login"
+
+  get 'auth/:provider/callback', to: 'users#authenticate', as: :authenticate
```
andela-ooranagwa: I think we also need a callback failure `route` or something :fried_shrimp:

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    token = @current_user.refresh_token
+    redirect_url = get_attrs_from_session([:redirect_url]).first
+
+    session.clear
+    redirect_to redirect_url
+    
+  end
+
+
+  def renew_token
```
andela-ooranagwa: Nice. 
Of course we expect that this route be called before the existing valid token expires.

Come to think of it, do you think there is any security implications in being able to refresh token at will :thought_balloon:

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  def tags
+    @user.subscribed_tags
+  end
+
+  private
+
+
+    def set_user
+      @user = User.find_by(id: params[:id])
+      resource_not_found and return unless @user
```
andela-ooranagwa: No sure though, but I think the use of `and` is *deprecated*

app/controllers/answers_controller.rb
andela-oakinniranye: No, the `question.answers` returns an active_record relation object which you can attach other AR queries to, like find, where, select etc.

app/controllers/answers_controller.rb
andela-oakinniranye: This has been resolved in the next commit

app/controllers/concerns/provider_helper.rb
andela-oakinniranye: Yeah, its not been caught. But I'd fix that in the next few pushes

app/controllers/application_controller.rb
```diff
@@ -1,14 +1,43 @@
...
     end
-    head 401, content_type: "application/json"
   end
+
+  def unauthorized_token
+    self.headers['WWW-Authenticate'] = 'Token realm="Application"'
+    render json: {errors: "Request was made with invalid token"}, status: 401
+  end
+
+  def resource_not_found
```
andela-ooranagwa: Nice thinking. I propose that all `404`'s be modified to use this

app/controllers/users_controller.rb
andela-ooranagwa: Given that `:resource_not_found` _renders_, and need for the `:return` statement?

app/controllers/users_controller.rb
andela-oakinniranye: :thought_balloon: left it as a comment so it could be referenced later. Would normally put these comments in a `.gitignored` file but thought others might need this as well

app/controllers/users_controller.rb
andela-oakinniranye: can't give a yes or no answer to this now, but if we can come up with a standard way that we'd use across the application, that would be fine

app/controllers/users_controller.rb
andela-oakinniranye: because programmers are lazy. useful when I was writing the code, because it seemed like a shorthand

app/controllers/users_controller.rb
andela-oakinniranye: Nice catch :100: :+1: , that would have been a problem ultimately, would fix that

app/models/user.rb
```diff
@@ -3,8 +3,25 @@ class User < ActiveRecord::Base
...
+    token
+  end
+
+  def self.from_token(token)
+    user_token = TokenManager.decode(token)
+    user = find_by(id: token[:user])
```
andela-ooranagwa: I think you meant to use `user_token` instead of `token` here.

app/controllers/users_controller.rb
andela-oakinniranye: :thought_balloon: food for thought. that reminds me, the current_user implementation ought to check if token has expired or does `JWT` execute that check for us?

app/controllers/users_controller.rb
andela-oakinniranye: `and` is a ruby method, I don't think it has been deprecated. but I think `&&` should work better since `and` has lower precedence. Would fix in the next push

app/models/user.rb
andela-oakinniranye: nope. but it would be useful if we want to attach more processes to the `refresh_token` method

config/routes.rb
andela-oakinniranye: sure, that's needed, would look into that

app/controllers/application_controller.rb
```diff
@@ -1,14 +1,43 @@
...
 
   def authenticate_token
-    payload, _header = TokenManager.authenticate(request)
-    if payload
-      @current_user = User.find_by(id: payload["user"])
-      return @current_user if @current_user && @current_user.active
+    authenticate_with_http_token do |auth_token, options|
+      payload, _header = TokenManager.decode(auth_token)
+      user_id = payload ? payload['user'] : nil
+      @current_user = User.find_by(id: user_id)
```
andela-ooranagwa: Nice rewrite. Cool.
We can also go further and modify `TokenManager.decode` method to return just the `payload` since we don't really need the header. 
Then we can modify this method to be something like:
```ruby
payload = TokenManager.decode(auth_token)
@current_user = User.find_by(id: payload["user']) if payload
```

config/routes.rb
```diff
@@ -13,15 +14,21 @@
...
+  match "*path", to: 'application#resource_not_found', via: :all
```
andela-ooranagwa: Nice :+1:

db/migrate/20160117215448_create_users.rb
```diff
@@ -1,6 +1,8 @@
 class CreateUsers < ActiveRecord::Migration
   def change
     create_table :users do |t|
+      t.string :name
```
andela-ooranagwa: Just thinking; is there a way to aggregate all user information in one table? `email`, `name` are in `users` table while, `profile_picture`, `profile_url`, `profile_email` are in `social_providers` :thought_balloon:

andela-oakinniranye: Yeah! That would have been preferred, but the problem is: `some users have 2 andela email addresses`. For example on my slack, I'm signing in using `oreoluwa.akinniranye@andela.co` while on google its `.com`, so we need a way of accounting for this, hence the reason for the two records. It would be fixed though, if we use regex to filter out `.co|.com` that should be solved, i think

app/controllers/users_controller.rb
andela-oakinniranye: yes, `:resource_not_found` renders, but the problem with Rails, is that it doesnt stop execution at the point of render, if you just call `render` without returning it would still try to execute the other commands. You may try this: after a `render` add a `puts "=-=--=-=-=-=-This is just a test-=-=-=-=-=--="` and visit the route, then check your logs, if that is outputted. `render` != `return`

docs/api.md
```diff
@@ -3,4 +3,4 @@
 The most important parameter for all endpoints except the `Home#index` is the
```
andela-ooranagwa: Any reason to allow passing token via params given the apparent security vulnerabilities?

app/controllers/users_controller.rb
andela-ooranagwa: I see. That's really thoughtful of you :raised_hands:

docs/api.md
andela-oakinniranye: absolutely, but current implementation doesn't support that, though as it can only be via the header. However, I think it may come down to the `FE` guys, i.e. what options does `react` provide?

app/controllers/answers_controller.rb
andela-ooranagwa: Oh, ok.
Please do verify; I agree the result is a collection of `AR` objects (given the `has_many` relation), but my thought is that it might return an empty collection.

app/controllers/application_controller.rb
andela-ooranagwa: :+1:

app/controllers/concerns/provider_helper.rb
andela-ooranagwa: Nice :+1:

app/controllers/users_controller.rb
andela-ooranagwa: lolz! Yes, but we still need provide a standard professional outlook

app/models/user.rb
andela-ooranagwa: :sunglasses:

app/controllers/users_controller.rb
andela-ooranagwa: Yea. JWT does the check; it's one of the errors (that might be) _thrown_ while decoding the token.

app/controllers/answers_controller.rb
andela-oakinniranye: Yes an empty `collection` can still be queried, the problem will occur when the question is empty, which the resource not found ought to take care of

app/controllers/users_controller.rb
andela-ooranagwa: Oh yes, I've notice that while debuging many `doubleRenderException` errors
:+1: :100:

andela-ooranagwa:
:+1:

andela-aadepoju:
Nice Work Sir Ore. :+1:

PR #13, opened 2016-01-31
Ft questions resources

andela-aadepoju:


app/models/question.rb
```diff
@@ -12,6 +12,10 @@ def with_answers
       includes(:answers)
     end
 
+    def top
+      where("votes > ?", 0).order("votes DESC")
```
andela-ooranagwa: Hmmm ... 
I don't think there is any need for the `where` condition we are more interested in the `ordering` of the votes e.g

> order("votes DESC")

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
+class QuestionsController < ApplicationController
+  before_action :set_question, only: [:show, :update, :destroy]
+
+  def index
+    questions = Question.all
+    render json: questions
```
andela-ooranagwa: Is the status implied?

app/models/question.rb
andela-aadepoju: :+1:

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  def show
+      render json: @question, status: 200
+  end
+
+  def create
+    question = Question.new(questions_params)
+    if question.save
+      render json: question, status: :created
+    else
+      render json:  question.errors, status: 400
```
andela-ooranagwa: @andela-oojewale made some suggestions on how best to return `form errors`

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def update
+    if @question.update(questions_params)
+      render json: @question, status: :ok
+    end
+  end
+
+  def destroy
+    if @question.try(:destroy)
+      render json: :no_content, status: 204
```
andela-ooranagwa: Using `head 204` might be shorter

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  def top_questions
+    questions = Question.top
+    render json: questions
```
andela-ooranagwa: Status missing

andela-aadepoju: This has been fixed

docs/questions_endpoints.md
```diff
@@ -196,10 +196,10 @@ Status: 404
   }
 ```
 
-## GET /questions/top_questions
+## GET /top_questions
 Request
 ```ruby
- GET  /questions/top_questions?limit=5,offset=1
+ GET  /top_questions?limit=5,offset=1
```
andela-ooranagwa: > GET  /questions/top_questions?limit=5&offset=1

app/controllers/questions_controller.rb
andela-aadepoju: Where did he make the suggestions again?

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      render json:  question.errors, status: 400
+    end
+  end
+
+  def update
+    if @question.update(questions_params)
+      render json: @question, status: :ok
+    end
+  end
```
andela-ooranagwa: Just wondering; what happens _if @question.update()_ is not successful? :thought_balloon:

andela-ooranagwa:
The alternatives to most of the `conditionals` are not provided; I believe that providing them might be relevant is some cases. 
Aside that (and the other minor comments I shared above), nice work chief. Cooool.

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    question = Question.new(questions_params)
+    if question.save
+      render json: question, status: :created
+    else
+      render json:  question.errors, status: 400
+    end
+  end
+
+  def update
+    if @question.update(questions_params)
```
andela-oakinniranye:
```ruby
if @question.update(question_params)
render json: @question, status: 200
else
render json: @question.errors, status: :unprocessable_entity
end
```
Handle the `conditions`, else the frontend may assume that the request was successful.

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  private
+
+  def set_question
+    @question = Question.find_by(id: params[:id])
+    resource_not_found && return unless @question
+  end
+
+  def questions_params
```
andela-oakinniranye: this should be `question_params` not `questions_params`

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    params.permit(:id, :title, :content, :votes, :user_id)
```
andela-oakinniranye: If we're authenticating a user via the `Authorization header`, we may not need to pass the `:user_id` params from the frontend anymore

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    else
+      invalid_request("field can't be blank")
+    end
+  end
+
+  def update
+    if @question.try(:update, questions_params)
+      render json: @question, status: 200
+    else
+       invalid_request("field can't be blank")
```
andela-oakinniranye: `invalid_request("field can't be blank")` how about when the update couldn't be done due to other errors. `Field can't be blank is a bit generic`

andela-oakinniranye:
:+1:

app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  def top_questions
+    questions = Question.top
+    render json: questions, status: 200
+  end
+
+  private
+
+  def set_question
+    @question = Question.find_by(id: questions_params[:id])
+    resource_not_found && return unless @question
```
andela-echigbo: Do we need the `return` statement? I think the `resource_not_found` method does  render of the response.

andela-aadepoju: The return is necessary when if we need to add anything after that line in future. It doesn't harm to leave it there now too.

app/controllers/application_controller.rb
```diff
@@ -1,17 +1,21 @@
...
   include ActionController::HttpAuthentication::Token::ControllerMethods
+  include ActionController::Serialization
+
 
   attr_reader :current_user
   before_action :authenticate_user
 
   def resource_not_found
-    render json: {errors: "The resource you tried to access was not found"}, status: 404
+    not_found = "The resource you tried to access was not found"
```
andela-ooranagwa: Nice! Keeping to style guide :+1:

PR #16, opened 2016-02-01
Added a module that checks if user owns the resource he's trying to modify

andela-oakinniranye:
A user should not be able to delete or update a resource he doesn't own, therefore this


app/controllers/answers_controller.rb
```diff
@@ -26,13 +28,16 @@ def update
...
 
   def destroy
-    @answer.destroy
-    head :no_content
+    if @answer.try(:destroy)
+      head :no_content
+    else
+      invalid_request(error_msg)
+    end
   end
```
andela-ooranagwa: Nice. I guess this skipped our attention earlier.

andela-ooranagwa:
This should be extended to the `questions` and `comments` *controllers*

andela-ooranagwa:
Awesome work :+1:

app/controllers/concerns/ownership_concern.rb
```diff
@@ -0,0 +1,27 @@
+
+require 'forwardable'
```
andela-ooranagwa: Could share more insight on this and what it is used for?

andela-oakinniranye: The ruby `forwardable` module allows you to delegate certain methods to other objects in your application.
Say you want to create a Queue system which uses an Array to queue items, one way to loop through would be to 
```ruby
def each(*item, &block)
queue.each{ |x|
 yield x if block_given?
}
end
```
However with the forwardable module, you could just delegate that method call to the appropriate object and let it handle it that way. So for example you could `def_delegator @queue, :each, :first, :last, :max, :min` and other array methods you would like to delegate to.
In my case `def_delegator self, :resource_name` which allowed me to call the class_method `resource_name` inside an instance_method

andela-oakinniranye:
Yes @andela-ooranagwa that's why I created it, so the other related controllers can use it too. Don't know how it would be useful to the `CommentsController`, since it's handling both the questions and the answers. It can be figured out though

andela-aadepoju:
:+1:

andela-aadepoju:
:+1:

app/controllers/answers_controller.rb
```diff
@@ -26,13 +28,16 @@ def update
...
-      render json: @answer.errors, status: :unprocessable_entity
+      invalid_request(error_msg)
     end
   end
 
   def destroy
-    @answer.destroy
-    head :no_content
+    if @answer.try(:destroy)
+      head :no_content
```
andela-echigbo: Is it not required to state the response `status` when the response has `no content` like this case and the `update` action?

andela-oojewale:
:+1:

app/controllers/answers_controller.rb
andela-oakinniranye: the `head :no_content` returns a response of `status: 204`. So it would not be required here

PR #23, opened 2016-02-05
Enhance Question Resource - issue 18

andela-echigbo:
Why?
To return a question's user object, votes_count, answers_count, views_count and time_updated


app/serializers/question_serializer.rb
```diff
@@ -1,8 +1,28 @@
...
 
-   has_many :comments, as: :comment_on
-   has_many :tags, as: :subscriber
-   has_many :answers
+  has_many :comments, as: :comment_on
+  has_many :tags, as: :subscriber
+  has_many :answers
+
+  def votes_count
+    object.votes.count
```
andela-oakinniranye: The `votes` are stored as either 1 or -1in the `value` column. What you may want is the sum(:value) not count. It's been fixed in my `PR`, though

app/serializers/question_serializer.rb
```diff
@@ -1,8 +1,28 @@
...
+    object.answers.count
+  end
+
+  def views_count
+    object.views
+  end
+
+  def user
+    {
+      name: object.user.name,
```
andela-oakinniranye: You should also include the user.id

app/controllers/questions_controller.rb
```diff
@@ -8,6 +8,7 @@ def index
   end
 
   def show
+    @question.increment_views
```
andela-oakinniranye: Instead of redefining the `increment_views` on the questions model. You can use the ActiveRecord `increment!` method, like so: `@question.increment!(:views, 1)`

app/serializers/question_serializer.rb
```diff
@@ -1,8 +1,28 @@
...
+
+  def votes_count
+    object.votes.count
+  end
+
+  def answers_count
+    object.answers.count
+  end
+
+  def views_count
```
andela-ooranagwa: Just thinking; is there any need for creating this as a separate method in the model? Given that it simply circles back to call _just_ one column in same model table :thought_balloon:

andela-oakinniranye: @andela-ooranagwa  this has been fixed in the `pull 22`

db/schema.rb
```diff
@@ -51,7 +52,7 @@
     t.string   "token"
     t.string   "profile_picture"
     t.string   "profile_url"
-    t.string   "email"
+    t.string   "profile_email"
```
andela-ooranagwa: Hmmm... the `migration` that causes this change doesn't seem to be included? :open_mouth:

andela-ooranagwa:
:+1:

andela-oakinniranye:
:+1:

app/serializers/question_serializer.rb
andela-echigbo: :100:

app/serializers/question_serializer.rb
andela-echigbo: Okay

app/controllers/questions_controller.rb
andela-echigbo: Interesting, I didn't know ActiveRecord has such method. Cool. I will change that

app/serializers/question_serializer.rb
andela-echigbo: :100: @andela-ooranagwa. @andela-oakinniranye does it mean you have implemented this feature already? Your `PR` has migration to add the `views` field?

db/schema.rb
andela-echigbo: I didn't change that man, trust me :open_hands:

PR #26, opened 2016-02-12
Update question_controller and question_serializer to return answers …

andela-ooranagwa:
…for show endpoint. Create User serializer to embed user information alongside api_key


andela-oakinniranye:
:+1:

PR #27, opened 2016-02-12
General update to serializers

andela-ooranagwa:
* Update serializers to use UserSerializer attribute definition for serializing user object
* Add user points to UserSerializer
* Embed user objects in Question answers
* Add default value, 0,  for user point in the db


andela-oakinniranye:
:+1:

PR #31, opened 2016-02-15
Refactor code

andela-oakinniranye:


app/serializers/answer_serializer.rb
```diff
@@ -1,13 +1,3 @@
...
-    {
-      id: object.user.id,
-      name: object.user.name,
-      email: object.user.points,
-      image: object.user.social_providers.first.try(:profile_picture),
-      points: object.user.points
-    }
-  end
+class AnswerSerializer < MainSerializer
+  attributes :question_id, :created_at, :updated_at, :comments_count
```
andela-ooranagwa: `:created_at` and `:updated_at` are already in *MainSerializer*, I don't think there is a need for them here.

app/serializers/comment_serializer.rb
```diff
@@ -1,5 +1,3 @@
-class CommentSerializer < ActiveModel::Serializer
-  attributes :id, :user, :content, :votes_count, :created_at, :updated_at
+class CommentSerializer < MainSerializer
```
andela-ooranagwa: Given that this is empty, we can delete it and have `comments` use `MainSerializer` directly.

app/serializers/answer_serializer.rb
andela-oakinniranye: :+1: nice observation

app/serializers/comment_serializer.rb
andela-oakinniranye: It's left there because we may have other comments we want to add to it later

app/serializers/main_serializer.rb
```diff
@@ -0,0 +1,17 @@
+class MainSerializer < ActiveModel::Serializer
+  attributes :id, :content, :votes_count, :voted, :created_at, :updated_at, :user
+
+  def voted
+    object.votes.voted?(object.class.to_s, object.id, object.user)
+  end
+
+  def user
```
andela-ooranagwa: Just a though; we can replace this with `belongs_to :user` to leverage the `UserSerializer`

config/routes.rb
```diff
@@ -1,10 +1,9 @@
 Rails.application.routes.draw do
-  post "questions/:question_id/upvote" => "votes#upvote"
-  post "answers/:answer_id/upvote" => "votes#upvote"
-  post "comments/:comment_id/upvote" => "votes#upvote"
-  post "questions/:question_id/downvote" => "votes#downvote"
-  post "answers/:answer_id/downvote" => "votes#downvote"
-  post "comments/:comment_id/downvote" => "votes#downvote"
+  scope '/:resource_name/:id', constraints: { resource_name: /(questions|answers|comments)/ } do
```
andela-ooranagwa: Nice clean up :+1:

andela-ooranagwa:
You'll might need to delete `common.rb` from *concerns* since it's no longer in user.

andela-ooranagwa:
Really impressive work as expected of a **Master**

:+1:

andela-oakinniranye:
`common.rb` is still being used in the `CommentsController`. Once we update that we can delete it completely. I guess @andela-oojewale would have to work on that.

andela-oojewale:
:+1:

PR #32, opened 2016-02-15
Add comment relationship to Question and Answer serializers

andela-ooranagwa:


andela-oakinniranye:
:+1: I think we may be needing the comments on the answers serializer as well @andela-ooranagwa

andela-ooranagwa:
Yes. I added to both the `Question` and `Answer` serializers.

PR #33, opened 2016-02-16
Ft refactor comments controller

andela-ooranagwa:


config/routes.rb
```diff
@@ -4,23 +4,22 @@
...
   end
 
   get "top_questions" => "questions#top_questions"
 
-  resources :answers, except: [:index, :show, :create, :destroy, :update, :new, :edit] do
-    resources :comments, except: [:new, :edit]
-  end
-
-  resources :comments, only: [:update, :delete]
+  resources :answers, except: [:index, :show, :create, :destroy, :update, :new, :edit]
```
andela-oakinniranye: If the 7 REST actions are excluded, we may not need the resources :answers here then

andela-ooranagwa: Nice catch. I ought to remove the entire resource. I've corrected this.

app/controllers/comments_controller.rb
```diff
@@ -1,76 +1,59 @@
...
+  def set_comment
+    @comment = @resource_comments.find_by(id: comment_params[:id])
+    resource_not_found && return unless @comment
 
-  def update_subject(id, user_id, attribute, value = nil)
-    if action_on_question
-      comments = Question.update_question_comment(id, user_id, question_id, attribute, value)
-    elsif action_on_answer
-      comments = Answer.update_answer_comment(id, user_id, answer_id, attribute, value)
+    unless (params[:action] == 'show') || (@comment.user.eql? current_user)
```
andela-oakinniranye: you may not really need this line. After the before_action :set_comment, just add `include OwnershipConcern` with that a user should be prevented from modifying a resource that isn't his

docs/comments_endpoints.md
```diff
@@ -4,8 +4,12 @@ Endpoints |	Usage |	Public Access
 --------- | ----- | -------------
 GET /answers/:id/comments |	Returns all the comments for a particular answer |	True
 POST /answers/:id/comments/ | Creates a new comment for the specified answer |	False
-PUT /comments/:id | Updates the comment with certain attributes | False
-DELETE /comments/:id | Deletes an comment | False
+PATCH /answers/:answer_id/comments/:id | Updates the specified comment of the specified answer | False
+DELETE /answers/:answer_id/comments/:id | Deletes the specified comment of the specified answer | False
+GET /questions/:id/comments |	Returns all the comments for a particular question |	True
+POST /questions/:id/comments/ | Creates a new comment for the specified question |	False
+PATCH /questions/:question_id/comment/:id | Updates the specified comment of the specified question | False
```
andela-oakinniranye: I think this route ought to be `PATCH /questions/:question_id/comments/:id` not comment

docs/comments_endpoints.md
```diff
@@ -72,11 +128,52 @@ Response
...
 {
-  message: "Successfully added"
+  id: 4,
+  text: 'This is such a scam',
+  date_created: 'Wed, 25TH Nov, 2017 12:00PM',
+  updated: true,
+  score: 9,
+  user: {
+    id: 3,
+    name: 'Bayo Owoade'
```
andela-oakinniranye: I don't know, but would it be possible to update all the docs to include all the json data we'd be returning. e.g for each of the resources I guess we'd be returning the user object with the profile :id, :name, :email, :image, :points etc.

app/controllers/comments_controller.rb
andela-ooranagwa: You are right. Nice catch.

docs/comments_endpoints.md
andela-ooranagwa: Noted. Corrected

docs/comments_endpoints.md
andela-ooranagwa: It's possible; but I'd suggest we leave that till we are done with the switch to `jbuilder` and then we could even use a standard API documentation tool.

andela-oakinniranye:
It would be great if we could use the same scope for the votes/upvotes and probably add a mixin to set_resource and perform other logics. Aside that :+1:

andela-ooranagwa:
@andela-oakinniranye could you clarify. I don't quite get you suggestion.

andela-oakinniranye:
like since we’re basically doing some things the same way in both  the VotesController and the CommentsController, we could nest the routes under the same scope and create a mixin that we could then include in both to set the resources

PR #34, opened 2016-02-18
Implement JBuilder for AnswersController

andela-ooranagwa:
* Add JBuilder gem
* Remove Active Model Serializer
* Generate JBuilder views and partials for Answers
* Implement JBuilder render in controller
* Create mock partials placeholders for users, default and comments JBuilder views


andela-oakinniranye:
:+1:

PR #35, opened 2016-02-21
Ft implement jbuilder for comment

andela-oojewale:


app/views/comments/index.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.array!(@resource_comments) do |comment|
+  json.partial! 'comments/default', data: comment
+  json.extract! comment, :comment_on_id, :comment_on_type
+  json.partial! 'users/user', user: comment.user
```
andela-echigbo: This `user` partial is already rendered in the `default` partial, won't this be a repetition?

app/views/comments/show.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.partial! 'comments/default', data: @comment
+json.extract! @comment, :comment_on_id, :comment_on_type
+json.partial! 'users/user', user: @comment.user
```
andela-echigbo: What I commented on the `index` view also applies here

app/views/comments/index.json.jbuilder
andela-oojewale: You're correct. I obviously did not pay attention to details there. Although, `jbuilder` would not render it twice. Corrected.

PR #36, opened 2016-02-22
Ft implement jbuilder for question controller

andela-oakinniranye:


app/models/question.rb
```diff
@@ -1,15 +1,17 @@
 class Question < ActiveRecord::Base
   include VotesCounter
+  include ActionView::Helpers::DateHelper
+
   has_many :comments, as: :comment_on
   has_many :votes, as: :voteable
-  has_many :tags, as: :subscriber
+  has_many :tags_questions
```
andela-echigbo: I notice here that you are spitting out what the `polymorphic` relationship does. Is there a need for that model being explicit? And this means the `User` model has to do the same

PR #39, opened 2016-02-27
Ft model test answer and user

andela-ooranagwa:
Create Factories for Resource models

A robust factory greatly eases implementing comprehensive test for the app, as such this _PR_:
* adds `factory_girl_rails` gem
* creates `factory` for the **User model**
* creates `factory` for the **Question model**
* creates `factory` for the **Answer model**
* creates `factory` for the **Comment model**
* creates `factory` for the **Vote model**
* creates `factory` for the **Tag model**
* adds `database_cleaner` gem


Gemfile
```diff
@@ -28,6 +28,8 @@ group :development, :test do
   gem "pry-rails"
   gem "pry-nav"
   gem "faker"
+  gem "factory_girl_rails"
```
andela-aadepoju: @andela-echigbo added this line too, maybe you should take a look at his branch so as to avoid a merge conflict.

andela-ooranagwa: Thanks @andela-aadepoju. We'll find a way to avoid the intending conflict.

PR #40, opened 2016-02-27
used eager loading and some other methods

andela-oakinniranye:


app/models/concerns/votes_counter.rb
```diff
@@ -16,7 +16,14 @@ module VotesCounter
   end
 
   def votes_count
-    try(:total_votes) || votes.sum(:value)
+    try(:total_votes) || votes_alternative
+  end
+
+  def votes_alternative
```
andela-ooranagwa: I think we can remove the `:to_i` and assignment to a variable (`t`). 
```ruby
def votes_alternative
  votes.reduce(0) do |_ , v|
        _ + v
     end
end
```
... and with variable name

app/models/question.rb
```diff
@@ -24,7 +24,9 @@ def time_updated
   end
 
   def self.with_associations
-    includes(:answers).includes(:user).includes(:comments)
+    eager_load(:votes).eager_load(answers: [{comments: [:user, :votes]}, :user, :votes]).
+    eager_load(:user).eager_load(comments: [:user, :votes]).eager_load(:tags)
+
   end
 
   def self.with_basic_association
```
andela-ooranagwa: Any need for this method?

app/models/question.rb
```diff
@@ -24,7 +24,9 @@ def time_updated
   end
 
   def self.with_associations
```
andela-ooranagwa: Nice :+1:

app/models/concerns/votes_counter.rb
andela-oakinniranye: yeah, that should be fixed

app/models/question.rb
andela-oakinniranye: yeah, one is useful for the `index` and the `show`

app/models/concerns/votes_counter.rb
```diff
@@ -16,7 +16,13 @@ module VotesCounter
...
+    votes.reduce(0) do |_ , v|
+      _.to_i + v.value
```
andela-ooranagwa: There is really no need for the conversion to integer (`:to_i`). What do you think?
... and @andela-iamadi will **wince** if he sees these variable names :speak_no_evil:

app/models/question.rb
andela-ooranagwa: I see. You are right.

PR #47, opened 2016-03-05
Ft model test vote and tag

andela-oojewale:


andela-aadepoju:
This tests are good to go. :+1:

PR #48, opened 2016-03-06
Implement the user reputation system.

andela-echigbo:
How?
When a question/answer/comment is upvoted, 5 points are added to the `points` attribute of the user that made the said resource.
If it is a downvote instead, 2 is subtracted from the that user
NB: If the same user upvotes and downvotes a resource, the previous option is canceled and the new one takes affect as though nothing happened.

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -1,6 +1,6 @@
 class AnswersController < ApplicationController
   before_action :set_question
-  before_action :set_answer, only: [:show, :update, :destroy]
+  before_action :set_answer, only: [:show, :update, :destroy, :accept]
   include OwnershipConcern
 
   def index
@@ -37,10 +37,16 @@ class AnswersController < ApplicationController
     end
   end
 
+  def accept
+    @answer.accept
+    render json: { message: "Answer Accepted" }, status: 201
+  end
+
   private
 
     def set_answer
-      @answer = @question.answers.find_by(id: params[:id])
+      id = params[:id] || params[:answer_id]
+      @answer = @question.answers.find_by(id: id)
       resource_not_found && return unless @answer
     end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,7 @@ class ApplicationController < ActionController::API
 private
   def authenticate_user
     authenticate_token || unauthorized_token
+    # @current_user = User.last
   end
 
   def authenticate_token
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,6 @@ class ApplicationController < ActionController::API
 private
   def authenticate_user
     authenticate_token || unauthorized_token
-    # @current_user = User.last
   end
 
   def authenticate_token
--- a/app/controllers/votes_controller.rb
+++ b/app/controllers/votes_controller.rb
@@ -1,20 +1,42 @@
 class VotesController < ApplicationController
+  before_action :check_user_voteable_status, :check_against_owned_resource
+
   def upvote
-    resource_to_upvote = vote_params[:resource_name].singularize.camelize.constantize
-    vote = Vote.act_on_vote('plus', resource_to_upvote, vote_params[:resource_id], current_user)
-    render json: { response: vote }, status: 200 unless vote.nil?
-    render json: { error: "Invalid vote!" }, status: 403 if vote.nil?
+    vote("plus")
   end
 
   def downvote
-    resource_to_upvote = vote_params[:resource_name].singularize.camelize.constantize
-    vote = Vote.act_on_vote('minus', resource_to_upvote, vote_params[:resource_id], current_user)
-    render json: { response: vote }, status: 200 unless vote.nil?
-    render json: { error: "Invalid vote!" }, status: 403 if vote.nil?
+    vote("minus")
   end
 
   private
+    def vote(vote_type)
+      vote = Vote.act_on_vote(
+        vote_type,
+        @resource_to_vote,
+        vote_params[:resource_id],
+        current_user
+      )
+      return invalid_request("Invalid vote!") if vote.nil?
+
+      render json: { response: vote }, status: 200
+    end
+
     def vote_params
       params.permit(:resource_name, :resource_id)
+    end
+
+    def check_user_voteable_status
+      invalid_request("Not qualified to vote") unless current_user.can_vote?
+    end
+
+    def check_against_owned_resource
+      @resource_to_vote = vote_params[:resource_name].singularize.camelize.constantize
+      resource = @resource_to_vote.find_by(id: vote_params[:resource_id])
+
+      return resource_not_found if resource.nil?
+
+      same_user = current_user.id == resource.user.id
+      invalid_request("You can't vote for your post") if same_user
     end
 end
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -27,6 +27,7 @@ class Answer < ActiveRecord::Base
                  },
                }
     )
+  end
 
   def accept
     self.accepted = true
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -1,4 +1,6 @@
 class Answer < ActiveRecord::Base
+  ACCEPTED_ANSWER_REWARD = 20
+
   include VotesCounter
 
   has_many :comments, as: :comment_on, dependent: :destroy
@@ -25,5 +27,10 @@ class Answer < ActiveRecord::Base
                  },
                }
     )
+
+  def accept
+    self.accepted = true
+    user.update_user_reputation(ACCEPTED_ANSWER_REWARD) if changed?
+    save
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,12 +33,12 @@ class User < ActiveRecord::Base
   end
 
   def can_vote?
-    return true # if points >= 15
-    false
+    points >= 15
   end
 
   def update_user_reputation(new_reward)
-    self.points += new_reward
-    save
+    new_point = points + new_reward
+    new_point = 0 if new_point < 0
+    update(points: new_point)
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,6 @@
 class User < ActiveRecord::Base
   include AndelaValidator
+
   has_many :comments
   has_many :questions
   has_many :answers
@@ -29,5 +30,15 @@ class User < ActiveRecord::Base
 
   def image
     social_providers.first.try(:profile_picture)
+  end
+
+  def can_vote?
+    return true # if points >= 15
+    false
+  end
+
+  def update_user_reputation(new_reward)
+    self.points += new_reward
+    save
   end
 end
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -3,6 +3,16 @@ class Vote < ActiveRecord::Base
   belongs_to :user
   validates :user, presence: true
 
+
+  REWARD = {
+    "question_upvote" => 5,
+    "answer_upvote" => 5,
+    "comment_upvote" => 2,
+    "question_downvote" => -2,
+    "answer_downvote" => -2,
+    "comment_downvote" => 0
+  }
+
   class << self
 
     def act_on_vote(operation, subject, subject_id, user)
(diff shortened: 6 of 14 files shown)


app/models/user.rb
```diff
@@ -30,4 +30,9 @@ def refresh_token
   def get_picture
     social_providers.first.try(:profile_picture)
   end
+
+  def can_vote?
+    return true # if points >= 15
```
andela-aadepoju: This will always return true or did you forget to uncomment `if points >= 15` ?

andela-echigbo: Thanks for that observation. It was actually intentional, as nobody currently has any point. The comment will just be uncommented once we have this sorted out. :wink:

app/controllers/answers_controller.rb
```diff
@@ -37,6 +37,11 @@ def destroy
     end
   end
 
+  def accept
```
andela-ooranagwa: We might want to also check that only the user that asked the question set an answer as accepted. Basically commend it to the cares of `Ownership concerns`

app/controllers/application_controller.rb
```diff
@@ -16,7 +16,8 @@ def invalid_request(message = error_msg)
 
 private
   def authenticate_user
-    authenticate_token || unauthorized_token
+    # authenticate_token || unauthorized_token
```
andela-ooranagwa: :see_no_evil:

app/models/answer.rb
```diff
@@ -1,5 +1,7 @@
 class Answer < ActiveRecord::Base
   include VotesCounter
+  include ReputationConcern
```
andela-ooranagwa: Just thinking that this concern might be better off in the `user` model given that all its operation/effects are reflect on a user object :thought_balloon:

app/controllers/answers_controller.rb
```diff
@@ -37,6 +37,11 @@ def destroy
...
+    Answer.accept(params[:answer_id])
```
andela-ooranagwa: Alternatively you could leverage the `before_action :set_answer` and then convert `:accept` to an _answer_ instance method

app/controllers/votes_controller.rb
```diff
@@ -7,6 +10,8 @@ def upvote
   end
 
   def downvote
+    return render json: { error: "Not qualified to vote" }, status: 403 unless current_user.can_vote?
```
andela-ooranagwa: Hmm; I see _WET_ code :bowtie:

app/models/answer.rb
```diff
@@ -11,4 +13,9 @@ class Answer < ActiveRecord::Base
   def self.with_associations
     includes(:user).includes(:comments).with_votes
   end
+
+  def self.accept(id)
+    Answer.find(id).update_attribute("accepted", true)
```
andela-ooranagwa: * I think `find` will work fine instead of `Answer.find` since we are in the **Answer** class
* Favor `find_by(:id, id)` instead of `find(id)` since the later could _blow up_ if `id` is `nil`

Please consider my case for `instance method` above :point_up:

app/models/user.rb
andela-ooranagwa: Points are also earned for _asking_ and/or _answering_ questions :smiley:

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,115 @@
...
+        it { expect(reward).to eq false }
+        it { expect { reward }.not_to change(Vote, :count) }
+      end
+
+      context "when previous is an upvote and current is a downvote" do
+        let(:reward) do
+          Vote.store_vote(vote.voteable.class, vote.voteable.id, user, -1)
+        end
+
+        it { expect(reward).to eq -7 }
```
andela-ooranagwa: Just curious, where is this **7** coming from?

app/controllers/answers_controller.rb
andela-echigbo: I have done that boss. I only added the `:accept` to the array of the before actions in the `OwnershipConcern`. So cool is reusable code. :recycle:

app/controllers/answers_controller.rb
andela-echigbo: The `accept` can not be an instance method, because it is calling the `update_user_reputation` class method. The later is made a class method because the method that calls it from the `Vote` model has been made a class method. ¯\_(ツ)_/¯

app/controllers/application_controller.rb
andela-echigbo: :speak_no_evil:

app/controllers/votes_controller.rb
andela-echigbo: :gift:

app/models/answer.rb
andela-echigbo: I think you are right. And then I will do `User.update_user_reputation` from both the `Vote.process_vote` and `Answer.accept` :bulb:

app/models/answer.rb
andela-echigbo: * I have removed the `Answer` :100: 
* Given that the upvote is clicked on an existing answer, I don't expect a case where the answer with the given `id` will not exist.

*Now that I have moved the concern to the `User` model, I think I can make it an instance method and treat your "case above :point_up: " :smile:

spec/models/vote_spec.rb
andela-echigbo: This should be clearer when you look at the recent commit :smile:

app/controllers/application_controller.rb
```diff
@@ -17,6 +17,7 @@ def invalid_request(message = error_msg)
 private
   def authenticate_user
     authenticate_token || unauthorized_token
+    # @current_user = User.last
```
andela-ooranagwa: We'll be happier if you delete this :wink:

app/controllers/votes_controller.rb
```diff
@@ -1,20 +1,42 @@
...
+      invalid_request("Not qualified to vote") unless current_user.can_vote?
+    end
+
+    def check_against_owned_resource
+      @resource_to_vote = vote_params[:resource_name].singularize.camelize.constantize
+      resource = @resource_to_vote.find_by(id: vote_params[:resource_id])
+
+      return resource_not_found if resource.nil?
+
+      same_user = current_user.id == resource.user.id
```
andela-ooranagwa: `current_user == resource.user` also works

app/controllers/votes_controller.rb
```diff
@@ -1,20 +1,42 @@
 class VotesController < ApplicationController
+  before_action :check_user_voteable_status, :check_against_owned_resource
```
andela-ooranagwa: :thought_balloon:

app/controllers/votes_controller.rb
```diff
@@ -1,20 +1,42 @@
...
     def vote_params
       params.permit(:resource_name, :resource_id)
     end
+
+    def check_user_voteable_status
+      invalid_request("Not qualified to vote") unless current_user.can_vote?
+    end
+
+    def check_against_owned_resource
+      @resource_to_vote = vote_params[:resource_name].singularize.camelize.constantize
```
andela-ooranagwa: :thought_balloon:

app/models/user.rb
andela-ooranagwa: Please uncomment the conditional.

andela-ooranagwa: What do you think of doing this:
```ruby
def can_vote?
  points >= 15
end
```

app/models/user.rb
```diff
@@ -30,4 +31,14 @@ def refresh_token
...
+    false
+  end
+
+  def update_user_reputation(new_reward)
+    self.points += new_reward
```
andela-ooranagwa: Any need for `self`, it's usage doesn't seem consistent?

app/models/vote.rb
```diff
@@ -3,6 +3,16 @@ class Vote < ActiveRecord::Base
   belongs_to :user
   validates :user, presence: true
 
+
+  REWARD = {
```
andela-ooranagwa: Nice :+1:

config/routes.rb
```diff
@@ -13,7 +13,9 @@
       get "popular_answers"
     end
 
-    resources :answers, except: [:new, :edit]
+    resources :answers, except: [:new, :edit] do
+      get "/accept", to: "answers#accept"
+    end
```
andela-ooranagwa: Here's an implementation I learnt from **BOSS** @andela-oakinniranye:
```ruby
resources :answers, except: [:new, :edit] do
  member do
    get "accept"
  end
end
```
:smile:

spec/models/user_spec.rb
```diff
@@ -195,4 +195,32 @@
...
+
+      it { expect(user.can_vote?).to be_truthy }
+    end
+
+    context "when the user has less than 15 points" do
+      before(:each) do
+        user.update_attribute(:points, 14)
+      end
+
+      # it { expect(user.can_vote?).to be_falsey }
```
andela-ooranagwa: :sunglasses:

spec/models/user_spec.rb
```diff
@@ -195,4 +195,32 @@
       expect(user.social_providers.first.provider).to eql provider
     end
   end
+
```
andela-ooranagwa: Awesome man! :clap:

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,172 @@
...
   describe ".process_vote" do
-   it "returns an integer value after vote is processed" do
-    allow(Vote).to receive(:subject_exists?).and_return (true)
-    allow(Vote).to receive(:store_vote).and_return (true)
-    expect(Vote.process_vote(subject_class, subject_id, user, value)).to be_an Integer
-    expect((Vote.process_vote(subject_class, subject_id, user, value).abs)).to eq(1)
-   end
+    it "returns an integer value after vote is processed" do
+      allow(Vote).to receive(:subject_exists?).and_return (true)
+      allow(Vote).to receive(:store_vote).and_return (7)
```
andela-ooranagwa: Still curious about this **7** cause I think since `:store_vote` return value isn't assigned, _stubbing_ with `nil` won't hurt :weary:

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,172 @@
...
+      expect((Vote.process_vote(subject_class, subject_id, user, value).abs)).to eq(1)
+    end
   end
 
   describe ".process_vote" do
-   it "returns nil" do
-    allow(Vote).to receive(:subject_exists?).and_return (nil)
-    expect(Vote.process_vote(subject_class, subject_id, user, value)).to be nil
-   end
+    it "returns nil" do
```
andela-ooranagwa: Why not merge the two `describe ".process_vote"` as such all their expectations will be together and easily managed.

spec/models/vote_spec.rb
andela-ooranagwa: The understanding for this **7** still eludes me :weary:

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,172 @@
...
+      allow(Vote).to receive(:subject_exists?).and_return (nil)
+      expect(Vote.process_vote(subject_class, subject_id, user, value)).to be nil
+    end
   end
 
   describe ".process_vote" do
```
andela-ooranagwa: :point_up:

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,172 @@
...
-   it "throws an error on invalid params" do
-    expect(Vote.process_vote(subject_class, invalid_subject_id, invalid_user, value)).to be nil
-   end
+    it "throws an error on invalid params" do
```
andela-ooranagwa: Hmm, this sure doesn't seem like "throw(ing) an error on invalid params" test

spec/models/vote_spec.rb
```diff
@@ -16,81 +16,172 @@
...
   end
 
   describe ".voted?" do
-   it "returns true if the user has voted" do
-    vote
-    expect(Vote.voted?(subject_class, subject_id, user)).to eq(true)
-    expect(Vote.voted?(subject_class, subject_id, user, value)).to eq(true)
-   end
+    it "returns true if the user has voted" do
+      vote
```
andela-ooranagwa: Maybe add a `falsy` expectation to show it also returns false if the user has not voted
```ruby
it "returns true if the user has voted" do
expect(Vote.voted?(subject_class, subject_id, user)).to eq(false)
vote
:
:
end

```

andela-ooranagwa:
:+1:
*...aside the comprehensive understanding of the test which keeps eluding me :disappointed:*

PR #49, opened 2016-03-06
Robust search

andela-oakinniranye:


app/models/comment.rb
```diff
@@ -9,4 +10,21 @@ class Comment < ActiveRecord::Base
...
     includes(:user).with_votes
   end
+
+  def question
+    if comment_on_type == 'Answer'
+      comment_on.question
+    else
+      comment_on
+    end
+  end
```
andela-ooranagwa: This method doesn't seem to be used anywhere. Is there any need for it.
Also modifying it's name (to, say, `:get_question`) might be desirable.

PR #52, opened 2016-03-09
add question model tests and some few other tests.

andela-aadepoju:
I added tests to cover questions model and one test on comment model. I also set up `CI` integration for `Circle CI` and `Coveralls`. I removed the `test` folder because it was making the test fail.


spec/models/comment_spec.rb
```diff
@@ -28,6 +28,12 @@
     it { is_expected.to belong_to(:comment_on) }
   end
 
+  describe "#with_associations" do
+    it "has associations" do
+      expect(Comment.with_associations).not_to be nil
```
andela-ooranagwa: I don't know if this actually checks loading of associations. What do you think of:
```ruby
expect(Comment.all.first.association(:user).loaded?).to be false
expect(Comment.with_associations.first.association(:user).loaded?).to be true
```
This might detect association better and help detect when resources are not being `eager_loaded`

spec/models/question_spec.rb
```diff
@@ -1,5 +1,49 @@
...
+      question.created_at = DateTime.yesterday
+      expect(question.time_updated).to eq "less than a minute ago"
+    end
+  end
+
+    describe "#tags_to_a" do
+      it "is tagged" do
+        expect(question.tags_to_a).to eq []
+      end
+    end
```
andela-ooranagwa: What do think; say we try this out with a question sample that has tags so as to check that it doesn't always return an empty array
 ```ruby
it 'returns empty array if question has no tag' do
  expect(question.tags.count).to eq 0
  expect(question.tags_to_a).to eq []
end

it "returns array contain tag names" do
   tag = create(:tag, name: "Zhishi")
   create(:question_resource_tag, tag: tag, question: question)
   expect(question.tags.count).to eq 1
   expect(question.tags_to_a).to eq ['Zhishi']
end
```
or something in that line

File: spec/models/question_spec.rb [hunk 3]
```diff
@@ -1,5 +1,49 @@
...
+
+    describe "#increment_views" do
+      it "increments views" do
+        expect(question.increment_views).to be true
```
andela-ooranagwa: Checking the actual view count increases by 1 might be also desirable

spec/models/question_spec.rb
```diff
@@ -1,5 +1,49 @@
...
+
+    describe "#with_associations" do
+      it "has associations" do
+        expect(Question.with_associations).not_to be nil
+      end
+    end
+
+    describe "#with_basic_association" do
+      it "has basic associations" do
+        expect(Question.with_basic_association).not_to be nil
```
andela-ooranagwa: :point_up_2:

andela-ooranagwa:
Nice work bud.

andela-aadepoju:
Noted! :100:

.coveralls.yml
```diff
@@ -0,0 +1 @@
+repo_token: li11mQP9V16CrLOrtGsWWn42f3LuqDr9C
```
andela-ooranagwa: :see_no_evil: Token and other authentication resources shouldn't be pushed to remote :speak_no_evil:

app/controllers/questions_controller.rb
```diff
@@ -1,5 +1,6 @@
 class QuestionsController < ApplicationController
   before_action :set_question, only: [:show, :update, :destroy]
+  include OwnershipConcern
```
andela-ooranagwa: :+1: Thanks; this eluded us for a while.

spec/models/question_spec.rb
```diff
@@ -1,5 +1,61 @@
 require 'rails_helper'
 
 RSpec.describe Question, type: :model do
+  subject { build(:question, user: user) }
+  let(:user) { build(:user) }
+  let(:question){create(:question, user: user)}
+
```
andela-ooranagwa: :thought_balloon: you don't seem to have need for defining both `subject` and `question`

spec/models/question_spec.rb
```diff
@@ -1,5 +1,61 @@
...
+    describe "#increment_views" do
+      it "increments views" do
+        question.increment_views
+        expect(0..10).to cover(question.views)
+      end
+    end
+
+    describe "#with_associations" do
+      it "has associations" do
+        subject.save
```
andela-ooranagwa: What do you think of remove this line and using just the `:question` created?

spec/models/question_spec.rb
```diff
@@ -1,5 +1,61 @@
...
+   tag = create(:tag, name: "Zhishi")
+   create(:question_resource_tag, tag: tag, taggable: question)
+   expect(question.tags.count).to eq 1
+   expect(question.tags_to_a).to eq ['Zhishi']
+end
+
+    describe "#increment_views" do
+      it "increments views" do
+        question.increment_views
+        expect(0..10).to cover(question.views)
```
andela-ooranagwa: Interesting, could you help me understand what `:cover` does?

PR #55, opened 2016-03-13
add token manager test

andela-aadepoju:


.coveralls.yml
```diff
@@ -0,0 +1 @@
+repo_token: K3Pomdav7dFeKalQEyIcFu1R9hoP007gg
```
andela-ooranagwa: Please don't expose tokens and other private resources publicly :sos:

andela-aadepoju: How else can we put the token in this repo? it is needed by `circle ci` to send the test reports to `coveralls`

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+class Object
```
andela-ooranagwa: Why modify the `Object` class?

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  def headers
+    {"Authorization": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.
+      eyJ1c2VyIjoxLCJleHAiOjE0NTc5NjEyMzh9.
+      PTyASPDQ2BzMmktgnc57YXHEVPPiKS5lEg9xz2CgsUQMmtbptF3tnhvJ7YSLG3oX1HClKh"}
+  end
+end
```
andela-ooranagwa: What do you think of making this token to be auto generated. This here is probably valid for ~24hrs (or thereabout); after which these tests would start failing.

Since (valid) token will be required for some other tests in future, I suggest you create a `method` in the support folder which would be executed to generated valid token to be used (and reused) for every other test.

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+RSpec.describe TokenManager do
+let(:manager){TokenManager}
```
andela-ooranagwa: Using subject might be better :thought_balloon:

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    it "decodes token" do
+      token = manager.generate_token(1)
+      expect(manager.decode(token)).to be_kind_of(Array)
+      expect(manager.decode(token).length).to eq 2
+    end
+  end
+
+  describe "#authenticate" do
+    it "authenticates token" do
+      expect(manager.authenticate(obj)).to eq [nil, 401]
```
andela-ooranagwa: :disappointed:

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    it "checks invalid token" do
+      expect(manager.token(obj)).to eq nil
```
andela-ooranagwa: Interesting! Does this actually mean that the token is invalid? :thought_balloon:

andela-ooranagwa:
@andela-aadepoju Most of test are only checking the length of the token generated. Could you extend it to test for say:
* That the token actually expires
* That the token encodes and decodes information correctly
* That when a token is retrieved successfully from the header information

spec/services/token_manager_spec.rb
andela-aadepoju: Getting the token is not an issue, getting an object with `headers` method is the reason for this.

.coveralls.yml
andela-ooranagwa: check this out @andela-aadepoju 
https://circleci.com/gh/andela-iamadi/zhishi-backend/edit#env-vars

spec/services/token_manager_spec.rb
andela-ooranagwa: Any random class will also server
```ruby
class RandomHeaderClass
  def initialize(token)
     @token = token
 end

 def headers
   "Authorization" => @token
 end
end
```

spec/services/token_manager_spec.rb
andela-aadepoju: Yes it is because the user is not valid

spec/services/token_manager_spec.rb
andela-aadepoju: Noted!

spec/services/token_manager_spec.rb
andela-ooranagwa: But given the implementation of the `:token` method
```ruby
   def token(request_obj)
      request_obj.headers["Authorization"]
    end
```
It should only return `nil` if value of the key `Authorization` is `nil`, so it doesn't actually deal with validity of token :thought_balloon:

PR #56, opened 2016-03-14
add token manager test

andela-aadepoju:


spec/rails_helper.rb
```diff
@@ -75,4 +75,16 @@
     with.test_framework :rspec
     with.library :rails
   end
+
+  class RequestObj
```
andela-ooranagwa: This might be better off being in a file in the support test folder

andela-aadepoju: I tried that, but it didn't work well as some methods didn't work correctly. An example is the `secrete`  method.

andela-ooranagwa: Cool you draw @andela-oakinniranye or @andela-oojewale attention to this. Or maybe I will try it out myself.
But in the spirit of keeping things `DRY` we really need to abstract this out given that token generation will be vital to feature testing.

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+    it 'generates a valid token' do
+      expect(token.split('.').count).to eq 3
+      expect{subject.decode(token)}.not_to raise_error
+    end
+
+    describe 'token expiration' do
+      it 'sets expiration time to 24hrs by default' do
+        decoded = subject.decode(token)
+        expect(decoded.first['exp']).to be_within(100).of(24.hours.from_now.to_i)
```
andela-oakinniranye: Is there the need for the `be_within` matcher, as against something providing a date_range and expect the expiration date to cover a date_range. An alternative would be time_travel to a date after the token validity period and expect it to raise an error.
If you'd go the time_travel route, there's a gem `delorean` that allows something similar to this

spec/services/token_manager_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+        expired = subject.generate_token(user.id, Time.now)
+        expect{subject.decode(expired)}.to raise_error JWT::ExpiredSignature
+        expect{subject.authenticate(expired)}.not_to raise_error
+        expect(subject.authenticate(expired)).to eql rescued_response
+      end
+    end
+
+    describe "arguments" do
+      it "accepts only one arguement" do
+        expect{subject.authenticate}.to raise_error(ArgumentError, 'wrong number of arguments (0 for 1)')
```
andela-oakinniranye: I think this may be too rigid, as I can imagine refactoring this sometime later and the `authenticate` method now has to take just one or no parameter.

andela-ooranagwa: Yes boss! That's the idea. My take is that tests should stand guard of your codes and make sure that no unintentional change in behaviour goes unnoticed. As such if the behaviour of a part of the app is modified (consciously), then the test should be updated with it's new watch.

spec/services/token_manager_spec.rb
andela-ooranagwa: Thanks for pointing out the `delorean` gem. It's awesome, especially given that writing tests that involve time are tricky.
Using it here, however, might seem like overflogging the issue. The need for the `be_within` matcher is to allow for any form system drag or delay while running the test which might affect the calculated expectation of the test.
I'm all for reducing the time to say `2` seconds; but I once experienced up to *72* seconds while experiment locally.

andela-oakinniranye: yeah sure, but the `be_within(100).of(24.hours.from_now.to_i)` doesn't seem so clear, as my first of many questions was what is `100`? and then if someone is thinking of the `within_` you would have expected a range, which isn't, as we're passing an integer in here

spec/services/token_manager_spec.rb
andela-oakinniranye: makes sense for the tests to be guarded, but making it too rigid may not be a good thing. i think it could have been better as `expect{subject.authenticate}.to raise_error(ArgumentError)`. **flexible interfaces**

PR #59, opened 2016-03-16
Meta info

andela-oakinniranye:


app/models/concerns/votes_counter.rb
```diff
@@ -3,11 +3,17 @@ module VotesCounter
 
   included do
     scope :with_votes, -> {
-      joins("LEFT JOIN votes ON votes.voteable_id = #{table_name}.id AND votes.voteable_type = '#{to_s}'").select("#{table_name}.*, SUM(votes.value) AS total_votes").group("#{table_name}.id")
+      joins("LEFT JOIN votes ON votes.voteable_id = #{table_name}.id AND " \
```
andela-iokonkwo: What do you think of using `AREL` here instead of raw sql. The reason why sql should be only used as a last resort is that the `SQL` of different sql databases varies. It will be very difficult to switch to a different database engine when you have a lot of sql queries scattered throughout your application. 

Check out [this blog post](https://www.viget.com/articles/composable-sql-queries-in-rails-using-arel) to learn how powerful  `AREL` can be. You may never need to write a single line of SQL again. Unless the query is just too complex.

app/models/user.rb
```diff
@@ -9,6 +9,12 @@ class User < ActiveRecord::Base
   has_many :votes
   has_many :resource_tags, as: :taggable
   has_many :tags, through: :resource_tags
+  scope :with_statistics, ->{
+    joins("LEFT JOIN questions ON questions.user_id = users.id LEFT JOIN " \
```
andela-iokonkwo: Same with above.

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def current_page
+      resource.current_page
+    end
+
+    def total_pages
+      resource.total_pages
+    end
+
+    def is_first_page?
```
andela-iokonkwo: Predicate methods do not need to start with `is_?`. Since there is a question mark, we know that it returns a boolean value. So this method could be refactored to `first_page?`

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
...
+      resource.current_page == 1
+    end
+
+    def is_last_page?
```
andela-iokonkwo: Same with above.

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
+class PaginationPresenter
```
andela-iokonkwo: Is this different from creating a decorator with draper?

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
...
+      page_number = resource.previous_page
+      page_url(page_number)
+    end
+
+    def next_page
+      page_number = resource.next_page
+      page_url(page_number)
+    end
+
+    def out_of_bounds?
```
andela-iokonkwo: We don't need to define this method since we are using `method_missing` to catch all missing method and send to resource object.

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
...
+      resource.out_of_bounds?
+    end
+
+    def offset
```
andela-iokonkwo: Same with above.

app/presenters/pagination_presenter.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def method_missing(method_name, *args, &block)
+      if resource.respond_to? method_name
+        resource.send(method_name, *args, &block)
+      else
+        super
+      end
+    end
+
+    def respond_to_missing?(method_name, *)
```
andela-iokonkwo: :+1: Good job implementing respond_to_missing? after creating method_missing method.

app/models/user.rb
andela-oakinniranye: The lord Ikem, Arel was initially used for this, but I had to change to raw SQL because of a particular limitation, but I'm still trying to find a way around though.
```ruby
    def query
      user_table.project(user_table[Arel.star], question_table[:id].count.as("questions_asked"),
      answers_table[:id].count.as("answers_given")).outer_join(question_table).
      on(user_table[:id].eq(question_table[:user_id])).outer_join(answers_table).
      on(user_table[:id].eq(answers_table[:user_id])).group(user_table[:id]).to_sql
    end
```
but using `find_by_sql` doesn't return an `ActiveRecord::Relation/Collection` therefore it was discarding all the associated `includes` so something like `User.with_associations.find_by_sql(query)` discards the associations defined and just runs the `find_by_sql`.

app/presenters/pagination_presenter.rb
andela-oakinniranye: DraperDecorator decorates the model objects, however it doesn't seem to do well with collections. Added to that is the fact that this method works with paginated collections.

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      5.times{ create(:user) }
+    end
+
+    it "gets basic user information" do
+      get :index, format: :json
+
+      expect(parsed_json).to have_key("users")
+
+      sample_user = parsed_json['users'].first
+      expect(sample_user['id']).to be
```
andela-iokonkwo: What are you testing here with `expect(sample_user['id']).to be`?
Also it makes sense to group our verification phase together. The 4 phases of testing are:
- Setup
- Execution
- Verification
- Teardown
In this order.
Check out [this blog post](http://www.codeaffine.com/2014/08/18/junit-in-a-nutshell-test-structure/)  and [this blog post](https://robots.thoughtbot.com/four-phase-test)for a better explanation of 4 phase test

spec/models/queries/statistics_query_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe Queries::StatisticsQuery do
+  describe "#call" do
+    it "calls the defined association" do
```
andela-iokonkwo: In this test, there is no explicit verification phase. There is a setup phase and execution phase but no verification. A better way you can go is to use test spies like this:
```ruby
 allow(subject).to receive(:questions_asked)
 allow(subject).to receive(:answers_given)
 allow(subject).to receive(:join_associations)

subject.call

 expect(subject).to have_received(:questions_asked)
 expect(subject).to have_received(:answers_given)
 expect(subject).to have_received(:join_associations)
```
Note that you don't need to use `described_class.new` in this case. `subject` is `described_class.new`

PR #60, opened 2016-03-19
Add the `accepted` attribute to the answer object

andela-echigbo:
Having added the accepted attribute to an answer object, we need to return it to the front end so they can do whatever they want with it. :smile:  I know it is the simplest `PR` you have ever reviewed. Yes... life can be that simple. :sunglasses: 

Well, I don't think it is still the simplest `PR` you have reviewed. Don't shoot me... keep it for @andela-ooranagwa

### DIFF
--- a/db/migrate/20160319110950_update_user_reputation_from_existing_votes.rb
+++ b/db/migrate/20160319110950_update_user_reputation_from_existing_votes.rb
@@ -1,0 +1,31 @@
+class UpdateUserReputationFromExistingVotes < ActiveRecord::Migration
+  def up
+	  User.find_each{|user| user.increment!(:points, 10)}
+
+	  Vote.all.each do |vote|
+      subject = vote.voteable_type.constantize
+      value = vote.value
+      reward = vote.evaluate_reward(true, value, subject)
+      subject_id = vote.voteable_id
+      user = subject.find_by(id: subject_id).user
+      user.update_user_reputation(reward) if reward
+    end
+
+    User.joins(:answers).where(answers: {accepted: true}).update_all("points = users.points + 20")
+  end
+
+	def down
+    User.joins(:answers).where(answers: {accepted: true}).update_all("points = users.points - 20")
+
+    Vote.all.each do |vote|
+      subject = vote.voteable_type.constantize
+      value = vote.value
+      reward = vote.evaluate_reward(true, value, subject)
+      subject_id = vote.voteable_id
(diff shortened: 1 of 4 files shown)


andela-oakinniranye:
:+1:

andela-echigbo:
What is @coveralls talking about? :question:

andela-ooranagwa:
:+1:

db/migrate/20160319110950_update_user_reputation_from_existing_votes.rb
```diff
@@ -0,0 +1,31 @@
+class UpdateUserReputationFromExistingVotes < ActiveRecord::Migration
```
andela-iokonkwo: @andela-echigbo Can you explain the importance of this migration?

andela-oakinniranye: we already have users on the platform, who have asked or answered or have upvoted question/answers/comments, however there is supposed to be a reward for this. Hence the reason we require this migration, especially to ensure consistency across all the environments(local, staging and production). @andela-iokonkwo

db/migrate/20160319110950_update_user_reputation_from_existing_votes.rb
```diff
@@ -0,0 +1,31 @@
...
+	  Vote.all.each do |vote|
+      subject = vote.voteable_type.constantize
+      value = vote.value
+      reward = vote.evaluate_reward(true, value, subject)
+      subject_id = vote.voteable_id
+      user = subject.find_by(id: subject_id).user
+      user.update_user_reputation(reward) if reward
+    end
+
+    Answer.all.select(&:accepted).map(&:user).each{|u| u.update_user_reputation(20)}
```
andela-iokonkwo: This is an expensive query. You are fetching all answers before selecting the accepted ones. Then you went ahead to select the user for each accepted answer thereby introducing `N + 1` query. Queries in Migration needs to be fast. I might even go as far as using raw sql.
```ruby
execute <<-SQL
  UPDATE users u
  SET u.points = u.points + 20, 
  FROM 
    (SELECT mu.id 
     FROM user mu 
     JOIN answers a
     ON a.user_id = mu.id
     WHERE a.accepted = true) o
WHERE 
    o.id = u.id
SQL
```
Try it out and let me know if it works. Well if it doesn't, you can make it work. :smile:

PR #62, opened 2016-03-23
changed the evaluate reward call to be called on the class not instan…

andela-oakinniranye:
…ce in the migration


andela-oojewale:
:+1:

frankie-ug:
:+1:

PR #63, opened 2016-03-23
staging environments added

andela-oakinniranye:


andela-oojewale:
:+1:

config/database.yml
```diff
@@ -23,3 +23,7 @@ test:
 production:
   <<: *default
   database: db/production.sqlite3
+
+staging:
+  <<: *default
+  database: db/staging.sqlite3
```
frankie-ug: are we using sqlite3 or postgresql?

andela-oakinniranye: locally we're using sqlite but on production we're using postgresql

frankie-ug: https://github.com/andela-iamadi/zhishi-backend/blob/master/config/database.yml#L25 seems to be pointing to sqlite3 or is there something I am missing?

andela-oakinniranye: by default rails would use whatever is defined as `ENV['DATABASE_URL']` otherwise it would revert to whatever is defined in the database.yml. Since we're planning for the scaling now and migrating to AWS, I guess we need to edit the config, but `ENV['DATABASE_URL']` is always defined on Heroku.
http://edgeguides.rubyonrails.org/configuring.html#configuring-a-database

frankie-ug: Just checked now. I think Heroku uses it's on config for production database. I think we still need to fix our database.yml for example https://github.com/andela-iamadi/zhishi-backend/blob/master/config/database.yml#L24 still uses the default configuration for production and our default is sqlite

frankie-ug: :+1:

andela-ooranagwa:
:+1:

PR #65, opened 2016-03-25
Ft tag resources

andela-oakinniranye:


andela-ooranagwa:
:+1:

app/controllers/answers_controller.rb
```diff
@@ -38,6 +38,7 @@ def destroy
   end
 
   def accept
+    return unauthorized_access unless @question.user == current_user
```
andela-echigbo: Awesome!!! :clap: :clap:

andela-echigbo:
:+1:

spec/models/tag_spec.rb
```diff
@@ -32,13 +32,62 @@
...
+
+    context "when other parameters are changed" do
+
+    end
+  end
+
+  describe "#update_tag_subscriptions" do
+
+  end
   # describe ".search" do
```
andela-echigbo: Though this was not done in this `PR`, but if it is not used, could you remove it? However, I think it will be nice to test the `.search` method, though. I think it's important.

spec/workers/tag_subscription_reassignment_worker_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe TagSubscriptionReassignmentWorker, type: :worker do
+  before do
+    allow(described_class).to receive(:perform_async)
+    3.times{ create(:tag, name: 'tag') }
```
andela-echigbo: Just an FYI, I don't know about the performance difference, but you could use `FactoryGirl`'s `create_list` like so:
```ruby
create_list(:tag, 3, name: 'tag')
```

PR #67, opened 2016-04-01
Ft show user custom questions

andela-ooranagwa:
* Add subscription_questions route
* Create subscription_questions action in QuestionsController
* Create from_user_subscription query methods in Question model
* Create custom_questions methods to DRY top_questions and subscription_questions
* Refactor top_questions and subscription_questions to use custom_question methods
* Update model tests; include test for custom questions methods

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -38,10 +38,12 @@ class QuestionsController < ApplicationController
     end
   end
 
+  def subscription_questions
+    custom_questions(:from_user_subscription, current_user.tags.pluck(:name))
+  end
+
   def top_questions
-    questions = Question.includes(user: [:social_providers]).top
-    @questions = PaginationPresenter.new(questions)
-    render :index
+    custom_questions(:top)
   end
 
   def search
@@ -57,5 +59,12 @@ class QuestionsController < ApplicationController
 
   def question_params
     params.permit(:title, :content)
+  end
+
+  def custom_questions(type, args=nil)
+    questions = Question.includes(user: [:social_providers])
+    questions = args ? questions.send(type, args) : questions.send(type)
+    @questions = PaginationPresenter.new(questions.paginate(page: params[:page]))
+    render :index
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/question.rb
```diff
@@ -34,6 +34,10 @@ def self.with_basic_association
     by_date.includes(user: [:social_providers])
   end
 
+  def self.from_user_subscription(user_tags)
+    joins(:tags).where(tags: { name: user_tags })
```
andela-oakinniranye: I think we may need to look into custom sql to fetch this, also I think we may prefer order by the tags in case there are no questions in the tags a user is subscribed to(http://stackoverflow.com/questions/866465/order-by-the-in-value-list). Scratch it, front end would take care of that: (when there are no questions in the user subscription, they should come to top questions). But we still need to use custom sql to avoid about 3 db calls here

app/controllers/questions_controller.rb
```diff
@@ -38,10 +38,12 @@ def destroy
     end
   end
 
+  def subscription_questions
+    custom_questions(:from_user_subscription, current_user.tags.pluck(:name))
```
andela-oakinniranye: same comments as the `question` model below

app/models/question.rb
andela-ooranagwa: My thought :thought_balloon:  on empty results would be to redirect the user to the _all questions route_. Arguable, if there shouldn't be a tag with at least one question (unless it's deleted), then, any user with empty result, in all likelyhood, have refused to subscribe to any tag.

Could you clarify on the 3 db calls you mentioned. Here's a sample query print out of the method:
```sql
SELECT "questions".* FROM "questions" INNER JOIN "resource_tags" ON "resource_tags"."taggable_id" = "questions"."id" AND "resource_tags"."taggable_type" = ? INNER JOIN "tags" ON "tags"."id" = "resource_tags"."tag_id" WHERE 1=0  [["taggable_type", "Question"]]
```
The only extraneous db query I seem to notice is that to retrieve the **user tag**. We could eliminate this by _side loading_ user tags when setting up **current_user**

PR #68, opened 2016-04-01
Add the user's vote on the resources and add test for the '.voted' method

andela-echigbo:
Why?
This is necessary so the frontend can to persist the user's vote on resources

How?
I added a `.voted` method to the `Vote` that takes the `user` and the `resource` instance and returns the `user`'s vote on that resource or `0` if the user has not voted for this resource.
I also made the `current_user` method a helper method so I can call it from the `JBuilder` view.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::API
   include ActionController::HttpAuthentication::Token::ControllerMethods
 
   attr_reader :current_user
+  helper_method :current_user
   before_action :authenticate_user
 
   def resource_not_found
--- a/app/models/concerns/votes_counter.rb
+++ b/app/models/concerns/votes_counter.rb
@@ -30,4 +30,9 @@ module VotesCounter
       sum + vote.value
     end
   end
+
+  def vote_by(user)
+    vote = votes.find { |vote| vote.user_id == user.id }
+    vote ? vote.value : nil
+  end
 end
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -82,11 +82,5 @@ class Vote < ActiveRecord::Base
       end
       reward
     end
-
-    def voted(resource, user)
-      vote = where(user_id: user.id, voteable_id: resource.id, voteable_type: resource.class.to_s).first
-      return vote.value if vote
-      0
-    end
   end
 end
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -82,5 +82,11 @@ class Vote < ActiveRecord::Base
       end
       reward
     end
+
+    def voted(resource, user)
+      vote = where(user_id: user.id, voteable_id: resource.id, voteable_type: resource.class.to_s).first
+      return vote.value if vote
+      0
+    end
   end
 end
--- a/app/views/answers/_answer.json.jbuilder
+++ b/app/views/answers/_answer.json.jbuilder
@@ -1,5 +1,5 @@
 json.extract! answer, :question_id, :comments_count, :id, :content, :votes_count, :accepted, :created_at, :updated_at
-json.voted Vote.voted(answer, current_user)
+json.user_vote answer.vote_by(current_user)
 json.user { json.partial! 'users/user', user: answer.user }
 json.comments(answer.comments) do |comment|
   json.partial! 'comments/comment', comment: comment
--- a/app/views/answers/_answer.json.jbuilder
+++ b/app/views/answers/_answer.json.jbuilder
@@ -1,4 +1,5 @@
 json.extract! answer, :question_id, :comments_count, :id, :content, :votes_count, :accepted, :created_at, :updated_at
+json.voted Vote.voted(answer, current_user)
 json.user { json.partial! 'users/user', user: answer.user }
 json.comments(answer.comments) do |comment|
   json.partial! 'comments/comment', comment: comment
--- a/app/views/questions/show.json.jbuilder
+++ b/app/views/questions/show.json.jbuilder
@@ -1,5 +1,6 @@
 json.partial! 'questions/question', question: @question
 json.partial! 'tags/tag', tags: @question.tags
+json.voted Vote.voted(@question, current_user)
 json.answers(@question.sort_answers) do |answer|
   json.partial! 'answers/answer', answer: answer
 end
--- a/app/views/questions/show.json.jbuilder
+++ b/app/views/questions/show.json.jbuilder
@@ -1,6 +1,6 @@
 json.partial! 'questions/question', question: @question
 json.partial! 'tags/tag', tags: @question.tags
-json.voted Vote.voted(@question, current_user)
+json.user_vote @question.vote_by(current_user)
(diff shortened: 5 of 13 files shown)


spec/factories/questions.rb
```diff
@@ -3,7 +3,7 @@
     title { Faker::StarWars.quote }
     content { Faker::Hipster.paragraph }
     views { rand(5) }
-    user
+    user factory: :user
```
andela-ooranagwa: I'm curious as why you want to do this :thought_balloon:

andela-ooranagwa:
Nice of you to update the documentation. We all need learn to do that :+1:

app/controllers/application_controller.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::API
   include ActionController::HttpAuthentication::Token::ControllerMethods
 
   attr_reader :current_user
+  helper_method :current_user
```
andela-ooranagwa: I see that you did this to enable `current_user` available in the view, right? We might have no need for the `attr_reader` then :thought_balloon:

app/models/vote.rb
```diff
@@ -82,5 +82,11 @@ def evaluate_reward(new_vote, value, subject)
       end
       reward
     end
+
+    def voted(resource, user)
+      vote = where(user_id: user.id, voteable_id: resource.id, voteable_type: resource.class.to_s).first
```
andela-ooranagwa: What do think of giving **votes** a default value of **0**, then we could just return the query result?
We can then modify this method to 
```ruby
select(:value).where(user: user, votable: resource).first
```
or something even better :wink:

app/views/answers/_answer.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 json.extract! answer, :question_id, :comments_count, :id, :content, :votes_count, :accepted, :created_at, :updated_at
+json.voted Vote.voted(answer, current_user)
```
andela-ooranagwa: Can we find a way of avoiding this query? Maybe with `eager_loading` or something? What you think @andela-oojewale, @andela-oakinniranye? :thought_balloon:

andela-ooranagwa:
Another :+1: for adding tests for the new feature you added.

app/views/answers/_answer.json.jbuilder
andela-oakinniranye: yeah. I don't know if it's better to define the voted method on the resource itself? so that we can do something like user.votes.where(voteable: resource) or even better: `has_many :voters, through: :votes, source: :user` and with that we could `resource.voters.include? user` on the model, we can then add the voters to the eager_loaded objects here. :thought_balloon:

db/schema.rb
```diff
@@ -76,7 +76,6 @@
     t.string   "name"
     t.datetime "created_at",        null: false
     t.datetime "updated_at",        null: false
-    t.integer  "parent_id"
```
andela-oakinniranye: nice catch

spec/factories/questions.rb
andela-echigbo: With this, when I create a question without passing a user object, it will build a user from the user factory and associate it to this question, and since a question belongs to a user, having a question without user association is not a good practice and will be disallowed in rails 5
`config.active_record.belongs_to_required_by_default = true`

app/controllers/application_controller.rb
andela-echigbo: We still need the `attr_reader` as it is what creates the `current_user` method, and the `helper_method` makes the just-created method a helper method. My thoughts, though.

app/models/vote.rb
andela-echigbo: But you know the user's vote record is only created when the user makes a vote. I think the query will return `nil` when no record of that match is found. Or does `select` work in a different way? 
cc @andela-ooranagwa, @andela-oakinniranye

db/schema.rb
andela-echigbo: Lol, which catch?? I only ran migration. :wink:

spec/factories/questions.rb
andela-ooranagwa: Now I'm even more curious, because the earlier implemenation automatically assigns a user if none is supplied. Could we pair on this or maybe [check this out](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#aliases); I seem to unable to reproduce your concern. Here's the output of the previous implemenation:
screen shot 2016-04-04 at 7 09 08 pm
It works as expected.

app/models/vote.rb
andela-ooranagwa: I see. You are right. Thanks for pointing that out.

app/models/vote.rb
```diff
@@ -82,5 +82,11 @@ def evaluate_reward(new_vote, value, subject)
...
+      return vote.value if vote
+      0
```
andela-ooranagwa: We can then do:
```ruby
vote.value || 0
```

andela-echigbo: This looks more sexy!! :smile:

andela-echigbo: Unfortunately, you I can't do this, because this will give me the vote object as a hash, whereas I want just the `value` field. :disappointed:

andela-echigbo: Unfortunately, you I can't do this, because this will give me the vote object as a hash, whereas I want just the `value` field. :disappointed:

spec/factories/questions.rb
andela-echigbo: I see @andela-oakinniranye. Probably something else was causing my problem then. I have removed it and it works fine.

app/views/answers/_answer.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 json.extract! answer, :question_id, :comments_count, :id, :content, :votes_count, :accepted, :created_at, :updated_at
+json.voted answer.vote_by(current_user)
```
andela-oakinniranye: I think you mean 
```ruby
{
...
vote: 0
...
}
```
not `voted`

app/models/concerns/votes_counter.rb
```diff
@@ -30,4 +30,9 @@ def votes_alternative
       sum + vote.value
     end
   end
+
+  def vote_by(user)
+    vote = votes.select { |vote| vote.user == user }.first
```
andela-oakinniranye: I think i've found the method we were looking for: `find`
You could now do
```ruby
  vote = votes.find{|vote| vote.user ==  user }
```
even better, if we don't want to add the `eager_load`, we could do: 
```ruby
  vote = votes.find{|vote| vote.user_id ==  user.id }
```

app/views/questions/show.json.jbuilder
```diff
@@ -1,5 +1,6 @@
 json.partial! 'questions/question', question: @question
 json.partial! 'tags/tag', tags: @question.tags
+json.voted @question.vote_by(current_user)
```
andela-oakinniranye:
```ruby
json.user_vote @question.vote_by(current_user)
```

andela-oakinniranye:
:+1:

PR #69, opened 2016-04-01
moved user image from social_provider to user model.

andela-oakinniranye:
There was a thought around removing the `image` || `profile_picture` from SocialProvider, but I thought since we'd still be getting rid of the social_providers table, that should still be left till we finally get rid of it

### DIFF
--- a/db/migrate/20160401195419_move_user_image_from_social_provider.rb
+++ b/db/migrate/20160401195419_move_user_image_from_social_provider.rb
@@ -1,0 +1,15 @@
+class MoveUserImageFromSocialProvider < ActiveRecord::Migration
+  def up
+    User.includes(:social_providers).find_each do |user|
+      image = user.social_providers.first.try(:profile_picture)
+      user.update(image: image)
+    end
+  end
+
+  def down
+    User.includes(:social_providers).find_each do |user|
+      image = user.image
+      user.social_providers.first.update(profile_picture: image)
+    end
+  end
+end
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -3,6 +3,7 @@ FactoryGirl.define do
     name { Faker::Name.name }
     points { Faker::Number.number(2) }
     sequence(:email) { |n| "#{Faker::Lorem.word}#{n}@andela.com" }
+    image { Faker::Avatar.image }
 
     factory :user_with_comments_on_question do
       transient do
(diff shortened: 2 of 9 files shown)


spec/models/user_spec.rb
```diff
@@ -111,7 +111,7 @@
 
     it "returns the url to user picture if user has picture" do
       image = Faker::Avatar.image
-      user.social_providers.create(profile_picture: image)
+      user.update(image: image)
```
andela-ooranagwa: Special :+1: :sunglasses:

andela-ooranagwa:
:+1:

spec/models/user_spec.rb
andela-ooranagwa: Couldl you also update the *User factory* to incude image:
```ruby
image { Faker::Avatar.image }
```

spec/models/tag_spec.rb
```diff
@@ -81,6 +81,24 @@
...
+    it "strips all whitespaces from the name" do
+      expect(subject.name).to eql(tag_name)
+    end
+  end
+
+  describe "#downcase!" do
+    let(:tag_name) { "RUBY-ON-RAILS"}
+    let(:subject) { create(:tag, name: tag_name)}
+
+    it "strips all whitespaces from the name" do
```
andela-oojewale: We might wanna say it `downcases the tag name` here and not `strip whitespaces`

andela-oojewale:
Don't know whats failing on coveralls though. Maybe somebody needs to look into that and *enlighten* me. That said, :+1:

spec/factories/users.rb
```diff
@@ -14,6 +15,10 @@
       end
     end
 
+    trait :without_image do
+      image nil
+    end
```
andela-echigbo: :+1:  This makes sense. Since it is not reused, another option would be making it an inner factory: 
```ruby
factory: :user_without_image { image nil }
```
and just create `user_without_image` factory

db/migrate/20160401195419_move_user_image_from_social_provider.rb
```diff
@@ -0,0 +1,15 @@
...
+    end
+  end
+
+  def down
+    User.includes(:social_providers).find_each do |user|
+      image = user.image
+      user.social_providers.first.update(profile_picture: image)
+    end
+  end
+end
```
andela-echigbo: :100: :icecream:

PR #70, opened 2016-04-03
Question having multiple accepted answers

andela-ooranagwa:
Fixes this be verifying all the answers to question and return error message if accepted answer already exists.

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -39,6 +39,7 @@ class AnswersController < ApplicationController
 
   def accept
     return unauthorized_access unless @question.user == current_user
+    return invalid_request('Question has an already accepted answer') if @question.answers.any?(&:accepted)
     @answer.accept
     render json: { message: "Answer Accepted" }, status: 201
   end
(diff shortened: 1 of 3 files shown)


app/controllers/answers_controller.rb
```diff
@@ -39,6 +39,7 @@ def destroy
 
   def accept
     return unauthorized_access unless @question.user == current_user
+    return invalid_request('Question has an already accepted answer') if @question.answers.any?(&:accepted)
```
andela-oakinniranye: do you think a `before_action :authorize_user, :validate_question_already_has_accepted_answer` works in this case?

andela-oojewale: I think that makes sense, the `before_action`.

andela-ooranagwa: A `before_action` makes a whole lot of sense. My thoughts however is that having a `before_action` for a **one line** method that is used just by one action is something of an _overkill_

andela-ooranagwa: I will be making it into a `before_action` nevertheless _inclusive of the other validation in the action_

PR #71, opened 2016-04-03
added a little helper to convert tag_params to array

andela-oakinniranye:


app/models/tag.rb
```diff
@@ -45,6 +44,16 @@ def resolution_for(tag_name)
         }
```
andela-oojewale: I can't seem to trace the opening counterparts of these braces. Please help.

andela-echigbo: Expand the contracted line blocks @andela-oojewale

andela-oojewale: Thanks!

andela-ooranagwa:
Can we also include a little helper test for this :stuck_out_tongue_winking_eye:

PR #72, opened 2016-04-04
Controller feature tests

andela-oakinniranye:


config/routes.rb
```diff
@@ -2,58 +2,58 @@
...
 
-  get "top_questions" => "questions#top_questions"
+    get "top_questions" => "questions#top_questions"
 
-  resources :users, only: [:show, :index] do
-    member do
-      get :questions
-      get :tags
-    end
+    resources :users, only: [:show, :index, :update] do
```
andela-ooranagwa: Yea but this; we were talking of removing the `update action` since it doesn't have any practical public use.

PR #74, opened 2016-04-05
Ft test answer requests

andela-ooranagwa:


spec/requests/answers/answer_request_helper.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+require "requests/shared/shared_authenticated_endpoint"
+
+def path_helper(path, answer=false)
+  question = FactoryGirl.create(:question_with_answers)
+  answer = FactoryGirl.create(:answer, question: question)
+  if answer
```
andela-oakinniranye: this method will always evaluate the first `if` block because even when answer is set to false, it's reassigned on line 6

andela-ooranagwa: Nice catch. Fixed

PR #75, opened 2016-04-06
Resource switcher

andela-oakinniranye:
This rake task enables the switching of a resource e.g from a comment to an answer or vice versa.

### DIFF
--- a/app/services/switch_resource_type.rb
+++ b/app/services/switch_resource_type.rb
@@ -5,11 +5,18 @@ module SwitchResourceType
       if from_record
         not_needed_attributes = send("#{from_record.class.table_name}_except")
         resource_attributes= from_record.as_json.except(*not_needed_attributes).merge(merge_attributes(from_record))
-        to_resource.create(resource_attributes)
+        to_record = to_resource.create(resource_attributes)
+        switch_votes(from: from_record, to: to_record)
         from_record.destroy
       else
         raise ActiveRecord::RecordNotFound
       end
+    end
+
+    def switch_votes(from:, to:)
+      new_voteable_type = to.model_name.to_s
+      new_voteable_id = to.id
+      from.votes.update_all(voteable_type: new_voteable_type, voteable_id: new_voteable_id)
     end
 
     def comments_except
--- a/app/services/switch_resource_type.rb
+++ b/app/services/switch_resource_type.rb
@@ -1,0 +1,47 @@
+class SwitchResourceType
+  def self.switch_resource(from_resource, to_resource, from_resource_id)
+    if from= from_resource.find(from_resource_id).destroy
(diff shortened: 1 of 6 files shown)


app/services/switch_resource_type.rb
```diff
@@ -0,0 +1,49 @@
...
+        resource_attributes= from.as_json.except(*not_needed_attributes).merge(merge_attributes(from))
+        to_resource.create(resource_attributes)
+      end
+    end
+
+    def comments_except
+      ["id", "comment_on_id", "comment_on_type"]
+    end
+
+    def self.answers_except
```
andela-echigbo: 👀  This is still within the `class << self` block

app/services/switch_resource_type.rb
```diff
@@ -0,0 +1,49 @@
...
+      elsif from.instance_of? Answer
+        {comment_on: question}
+      end
+    end
+
+    def permitted_resources
+      [ Comment, Answer ]
+    end
+
+    def method_missing(method_name, *args, &block)
```
andela-ooranagwa: `&block`

app/services/switch_resource_type.rb
```diff
@@ -0,0 +1,49 @@
+class SwitchResourceType
+  class << self
+    def switch_resource(from_resource, to_resource, from_resource_id)
+      if from= from_resource.find(from_resource_id).destroy
```
andela-ooranagwa: Can we have the assignment _cum_ argument to *if* evaluated on a separate line then you might want to consider when `resource` is not found :thought_balloon:

app/services/switch_resource_type.rb
```diff
@@ -0,0 +1,51 @@
...
+        from_resource, to_resource = [resource_match[:from_resource], resource_match[:to_resource]].map do |resource_type|
+          resource_type.singularize.camelize.constantize rescue nil
+        end
+        raise ResourceSwitchViolation unless ([from_resource, to_resource] & permitted_resources).size == permitted_resources.size
+        switch_resource(from_resource, to_resource, args[0])
+      end
+    end
+
+    def respond_to_missing?(method_name, *args)
+      return true unless method_name.match(/\A(?:answer|comment)_to_(?:answer|comment)\z/)
```
andela-toduah: Is this supposed to return true `unless` or `if`?

app/services/switch_resource_type.rb
andela-oakinniranye: ? :thought_balloon:

app/services/switch_resource_type.rb
andela-oakinniranye: nice catch. Fixed in the next push

PR #76, opened 2016-04-06
Add the CI badges to the README.md file.

andela-echigbo:
Add the CI badges to the readme.md file.

Note: I did not add the test coverage from `Code Climate`. Don't think it makes sense to have two the same badge


andela-oakinniranye:
:+1:

PR #77, opened 2016-04-07
Fixed fixable CodeClimate issues

andela-toduah:


app/controllers/concerns/provider_helper.rb
```diff
@@ -1,13 +1,8 @@
...
-    when /^slack/
-      '/auth/slack'
-    else
-      raise AuthProviderError.new("The provider is not supported on this platform")
-    end
-    provider_url
+    return '/auth/google_oauth2' if provider =~ /^google/
+    return '/auth/slack' if provider =~ /^slack/
+      
+    raise AuthProviderError.new("The provider is not supported on this platform")
```
andela-echigbo: Nice!! 🍺

app/controllers/users_controller.rb
```diff
@@ -15,7 +15,7 @@ def show
   def update
     if @user.update(user_params)
       render :show
-    else @user.errors
+    elsif @user.errors
       render json: @user.errors, status: :unprocessable_entity
     end
```
andela-echigbo: Could there be a case where the update is unsuccessful and yet no error? If so, then we should have an `else` to handle that case. If not, then I think we should just change the `elsif @user.errors` to `else`.

andela-toduah: That's true. I've implemented the else condition.

andela-oakinniranye:
:+1:

PR #78, opened 2016-04-10
Ft tag token request test

andela-oojewale:


app/controllers/tokens_controller.rb
```diff
@@ -1,10 +1,11 @@
-class TokensController < ApplicationController
+
+ class TokensController < ApplicationController
   before_action :set_token, only: [:validate]
   skip_before_action :authenticate_user, only: [:validate]
 
   def validate
     @user = @token.user
-    @user.update(active: true)
+    render json: {}, status: 200 if @user.update(active: true)
```
andela-oakinniranye: there's a jBuilder view that renders the json object for this, I guess. calling `render json: {}` just renders an empty object, whereas it ought to return the user's detail with token

db/schema.rb
```diff
@@ -17,6 +17,7 @@
     t.integer  "user_id"
     t.integer  "question_id"
     t.string   "content"
+    t.integer  "votes"
```
andela-oakinniranye: There's a votes table that handles the votes so we can easily manage the votes, I doubt we'd be needing this. same for `comments and questions`

spec/models/tag_spec.rb
```diff
@@ -4,9 +4,7 @@
   let(:arg) { :recent }
   let(:tag_name) { "Amity" }
   before :each do
-    create_list(:tag, 5, name: "Contract", created_at: 4.days.ago)
-    create_list(:tag, 3, name: "Amity", created_at: 2.days.ago)
-    create_list(:tag, 2, name: "Kaizen")
+    create_tags
```
andela-oakinniranye: where was this method defined? 💭

spec/requests/tags_request_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+
+RSpec.describe TagsController, type: :request do
+
+  let(:user) { valid_user }
+  let(:contact_tag) { Tag.find_by(name: "contract") }
```
andela-oakinniranye: i think this ought to be `contract_tag`

spec/requests/tags_request_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  before :each do
+    create_tags
+  end
+
+  describe "GET /tags" do
+    it "returns 200 status and tag" do
+      allow(Tag).to receive(:search).and_return([contact_tag])
+      get tags_path("contract"), {format: :json}, authorization_header
```
andela-oakinniranye: I think you mean: 
```ruby
get tags_path(q: 'contract'), {format: :json}, authorization_header
```

spec/requests/tags_request_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      get popular_tags_path, {format: :json}, authorization_header
+
+      expect(parsed_json).to have_key("contract")
+      expect(status).to eq 200
+    end
+  end
+
+  describe "GET /tags/recent" do
+    it "returns 200 status and recent tags" do
```
andela-oakinniranye: I think the block names should be more explicit. something like `returns status 200 with all tags ordered by the recently created ones`

spec/support/json_parser_helper.rb
```diff
@@ -6,6 +6,10 @@ def parsed_json
   def body
     response.body
   end
+
+  def status
+    response.status
+  end
```
andela-oakinniranye: nice 👍

spec/support/create_resource.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def create_user
+    create(:user)
+  end
+end
+
+RSpec.configure do |config|
+  config.include CreateResource
+end
```
andela-oakinniranye: nice work here 👍

spec/requests/tags_request_spec.rb
andela-oojewale: That defaults to the `q` though. I correct if for clarity.

spec/requests/tags_request_spec.rb
andela-oojewale: True. Thanks

spec/requests/tags_request_spec.rb
andela-oojewale: Cool

spec/support/json_parser_helper.rb
andela-oojewale: Thanks.

spec/support/create_resource.rb
andela-oojewale: Thanks

spec/requests/tags_request_spec.rb
andela-oojewale: I just found out that was wrong. Pardon me.

spec/rails_helper.rb
```diff
@@ -12,8 +12,8 @@
 require "shoulda/matchers"
 require "spec_helper"
 require "rspec/rails"
-require 'coveralls'
-Coveralls.wear!
+# require 'coveralls'
+# Coveralls.wear!
```
andela-echigbo: I think you should uncomment this, to give report to `coveralls`

app/controllers/tags_controller.rb
```diff
@@ -18,5 +18,6 @@ def trending
 
   def update_subscription
     current_user.tags = Tag.process_tags(params[:tags])
+    render json: {}, status: 200
```
andela-oakinniranye: I know it's not meant to be your task, but could you help write a jbuilder to return all the tags the user is now subscribed to?

PR #79, opened 2016-04-11
Ft test question requests

andela-echigbo:
Wrote tests for the question request endpoints

### DIFF
--- a/spec/models/vote_spec.rb
+++ b/spec/models/vote_spec.rb
@@ -60,20 +60,6 @@ RSpec.describe Vote, type: :model do
     end
   end
 
-  describe ".voted?" do
-    it "returns false if the user has not voted" do
-      expect(Vote.voted?(subject_class, subject_id, user)).to eq(false)
-      expect(Vote.voted?(subject_class, subject_id, user, invalid_value)).to eq(false)
-      expect(Vote.voted?(subject_class, invalid_subject_id, user, value)).to eq(false)
-    end
-
-    it "returns true if the user has voted" do
-      vote
-      expect(Vote.voted?(subject_class, subject_id, user)).to eq(true)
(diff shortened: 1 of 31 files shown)


spec/models/vote_spec.rb
```diff
@@ -100,7 +86,7 @@
     end
 
     context "when the user has not voted before" do
-      before(:each) { allow(Vote).to receive(:voted?).and_return (false) }
+      before(:each) { allow_any_instance_of(subject_class).to receive(:vote_by).and_return (nil) }
```
andela-iokonkwo: Just wondering. Why do we need to stub the vote_by method?. `allow_any_instance_of(:Class)` should be used with caution.

spec/requests/question/create_question_endpoint_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+          it { expect(response).to match_response_schema('question/show') }
+        end
+      end
+
+      context "with invalid question params" do
+        let(:question_params) do
+          [
+            { title: "This question has only a title", format: :json },
+            { content: "A question with no title", format: :json }
+          ].sample
```
andela-iokonkwo: I can't seem to find the place you used `question_params` you defined using let method.

andela-echigbo: This and other `question_params` defined in other context groups are used in the `before each` block (defined on line 11 as all the examples in all the context blocks needs this) to provide a context-specific `question_params` based on the given context.

spec/models/vote_spec.rb
andela-echigbo: You are right. I was trying to substitute the stub that was there. I don't really need to stub the `vote_by` method since it will still end up returning `nil` when the user has not made a vote. Thank you. 🙌  @andela-iokonkwo

andela-oakinniranye:
👍

PR #80, opened 2016-04-20
integrated bugsnag

andela-oakinniranye:


andela-echigbo:
@andela-oakinniranye, Could you please provide some description on what and how the `Bugsnag` works in the description area, and how we need it?

andela-oojewale:
@andela-echigbo, you can check it out [here](https://github.com/bugsnag/bugsnag-ruby).

PR #81, opened 2016-04-20
Unify tag endpoint response type

andela-ooranagwa:
The tag endpoints, recent, trending, popular, index, update_subscription return differing response object which can be a pain to handle. Some return an object, a couple return an array and some just the status code.
This implements a common response structure accross all the endpoint; an object with tags as key and an array of tag objects (containing tag :name, :id and :representative_id) as values. This is achieved by doing the following:

* modify tag queries to return a serializable collection
* create a new partial, tag_properties to serialize the collection as desired
* replace all former serializers with the new one
* Update the test and JSON scheme files

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -1,22 +1,22 @@
 class TagsController < ApplicationController
-
   def index
-    @tags =  Tag.search(params[:q])
+    render partial: 'tag_properties', locals: { tags: Tag.search(params[:q])}
   end
 
   def popular
-    render json: Tag.get_tags_that_are(:popular), status: 200
+    render partial: 'tag_properties', locals: { tags: Tag.get_tags_that_are(:popular) }
   end
 
   def recent
-    render json: Tag.get_tags_that_are(:recent), status: 200
+    render partial: 'tag_properties', locals: { tags: Tag.get_tags_that_are(:recent) }
   end
 
   def trending
-    render json: Tag.get_tags_that_are(:trending), status: 200
+    render partial: 'tag_properties', locals: { tags: Tag.get_tags_that_are(:trending) }
   end
 
   def update_subscription
     current_user.tags = Tag.process_tags(params[:tags])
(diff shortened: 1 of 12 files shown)


andela-oakinniranye:
👍

andela-oakinniranye:
Would it make sense to also change the `tags` property of the questions to also use the same `tags_properties` partial?

app/controllers/tags_controller.rb
```diff
@@ -1,22 +1,16 @@
...
-    @tags =  Tag.search(params[:q])
-  end
-
-  def popular
-    render json: Tag.get_tags_that_are(:popular), status: 200
-  end
-
-  def recent
-    render json: Tag.get_tags_that_are(:recent), status: 200
+    render partial: 'tag_properties', locals: { tags: Tag.search(params[:q])}
```
andela-echigbo: Is there a reason we can't use the `_tags` partial we already have or did you rename it to `tag_properties`?

My second suggestion would be, do you think it will be more readable and more conventional to set the `@tags` instance variable and then just render partial `tag_properties`, instead of passing `locals` hash which seem to be more low-level.  😨

andela-ooranagwa: @andela-echigbo a cursory look at the two partial will reveal that they serialise their payload differently and both partials _by extension the serialisation they offer_ are currently actively required _(at the least pending the [resolution of this suggestion](https://andela.slack.com/archives/zhishi-engine/p1461227332000005) by @andela-oakinniranye)_

andela-ooranagwa: Could you shed more light on the **low-level** nature of the use of `locals`? The [render documentation](http://apidock.com/rails/ActionController/Base/render) is not clear it's possibly non-conventional use.

Using an instance variable would be a great idea. But it's not as if the information is used in multiple places, just one. Do you think it's worth littering the namespace with instance variable(s) for lack of a better way to pass information from controller to view?

PR #82, opened 2016-04-21
Added Personalized Questions Route and Implementation

andela-oakinniranye:
As the application is growing, in order for users to engage with the platform more, they would like to see questions/topics personalized to their interests.

This PR ensures that once a user has subscribed to a tag or tags, when they come to a personalized question page they can see all questions personalized to their interest.
This addition should also be useful in the newsletter which would need to carry content personalized to the user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ association_count.rb
 /coverage/*
 /.coveralls.yml
 .DS_Store
+/.backup
+.codesnip
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -7,6 +7,12 @@ class QuestionsController < ApplicationController
     @questions = PaginationPresenter.new(questions)
   end
 
+  def personalized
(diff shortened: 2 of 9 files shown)


andela-ooranagwa:
@andela-oakinniranye this seems to be the same as #67

andela-echigbo:
I was thinking the idea was to make the `personalization` thingy the default behaviour of `Zhishi`, which means implementing it in the index action.

Also, could we make it an ordering process, where it still gets the pages of all `questions`, but order them by the `most-interest` first. That way, there will still be questions to display even if the user has not subscribed to any tag to personalise his views with.

andela-oakinniranye:
@andela-echigbo great thought. I don't think we agreed on having the personalized as default(index) action. But definitely something worth exploring. Would you mind adding this to the Trello board. Some sorting considerations we have to make would include the question ordering, time posted, activities(views, votes) etc. I think this may be useful: https://www.quora.com/What-are-the-best-practices-for-building-something-like-a-News-Feed

PR #83, opened 2016-04-21
Votes controller test

andela-aadepoju:
This PR adds test cases for votes controller.

@andela-echigbo @frankie-ug @andela-ooranagwa @andela-iokonkwo @andela-oakinniranye please review.

### DIFF
--- a/spec/controllers/votes_controller_spec.rb
+++ b/spec/controllers/votes_controller_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe VotesController, type: :controller do
 
   context "when user votes another user's resource" do
     before do
-      request.headers['Authorization'] = "Token token=#{user2.refresh_token}"
+      request.headers['Authorization'] = "Token token=#{valid_user_token(user2)}"
     end
 
     it "upvotes another user's resource" do
@@ -29,7 +29,7 @@ RSpec.describe VotesController, type: :controller do
 
   context "when user votes own resource" do
     before do
-      request.headers['Authorization'] = "Token token=#{user2.refresh_token}"
+      request.headers['Authorization'] = "Token token=#{valid_user_token(user2)}"
     end
 
     it "doesnt downvote resource" do
@@ -47,7 +47,7 @@ RSpec.describe VotesController, type: :controller do
 
   context "when user cannot vote on resource" do
     before do
-      request.headers['Authorization'] = "Token token=#{user3.refresh_token}"
+      request.headers['Authorization'] = "Token token=#{valid_user_token(user3)}"
     end
 
     it "doesnt upvotes another user's resource" do
--- a/spec/controllers/votes_controller_spec.rb
+++ b/spec/controllers/votes_controller_spec.rb
@@ -1,4 +1,86 @@
 require 'rails_helper'
 
 RSpec.describe VotesController, type: :controller do
+    let(:user1) { create(:user_with_voteable_status) }
+    let(:user2) { create(:user_with_voteable_status) }
+    let(:user3) { create(:user_without_voteable_status) }
+    before do
+      create_list(:question, 5, user: user1)
+      create_list(:answer, 5, user: user2)
+    end
+
+  context "when user votes another user's resource" do
+    before do
+      request.headers['Authorization'] = "Token token=#{user2.refresh_token}"
+    end
+
+    it "upvotes another user's resource" do
+      post :upvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to eq "{\"response\":1}"
+      expect(response.status).to eq 200
+    end
+
+    it "downvotes another user's resource " do
+      post :downvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to eq "{\"response\":-1}"
+      expect(response.status).to eq 200
+    end
+
+
+  end
+
+  context "when user votes own resource" do
+    before do
+      request.headers['Authorization'] = "Token token=#{user2.refresh_token}"
+    end
+
+    it "doesnt downvote resource" do
+      post :downvote, {resource_name: 'answers', resource_id: 1}
+      expect(response.body).to include "You can't vote for your post"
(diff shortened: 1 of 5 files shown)


spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,75 @@
...
+  end
+
+  context "when user downvotes another user's resource" do
+    before do
+      request.headers['Authorization'] = "Token token=#{user1.refresh_token}"
+    end
+
+    it "downvotes another user's resource" do
+      post :downvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to include "You can't vote for your post"
```
andela-echigbo: The context is saying: "when a user downvotes **another user's** resource", but the expected response is saying **your post**

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,75 @@
...
+    end
+
+    it "doesnt downvote user's own resource" do
+      post :downvote, {resource_name: 'answers', resource_id: 1}
+      expect(response.body).to eq "{\"response\":-1}"
```
andela-echigbo: "It doesn't downvote user's own resource", could you please explain how the response is expected to be `-1`?

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,75 @@
...
+      request.headers['Authorization'] = "Token token=#{user3.refresh_token}"
+    end
+
+    it "doesnt upvotes another user's resource" do
+      post :upvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to include "Not qualified to vote"
+    end
+
+    it "doesnt downvote user's own resource" do
+      post :downvote, {resource_name: 'answers', resource_id: 1}
```
andela-echigbo: Just curious... I don't think `user3` is the owner of `answer 1`?

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,75 @@
...
+      expect(response.body).to include "Not qualified to vote"
+    end
+  end
+
+  context "when user tries to vote with invalid token" do
+    before do
+      request.headers['Authorization'] = "Token token=badtoken"
```
andela-echigbo: Nice! 👍

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,75 @@
...
+    end
+
+    it "rejects invalid token for upvote" do
+      post :upvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to include "invalid token"
+    end
+
+    it "rejects invalid token for downvote" do
+      post :downvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to include "invalid token"
```
andela-echigbo: What do you think about using status codes here, as the response text might be changed?

spec/factories/users.rb
```diff
@@ -9,6 +9,14 @@
       active true
     end
 
+    factory :user_with_voteable_status do
+      points 15
```
andela-echigbo: Sweet! I like that you just gave him `15` only. No wasting of our points. We are not that generous. 😄 ¯\_(ツ)_/¯

spec/controllers/votes_controller_spec.rb
andela-aadepoju: Oh! That was an oversight. I think I have it all corrected. I will push the changes now.

andela-aadepoju:
@andela-echigbo: Thanks for your observations. I noticed all these myself and corrected them, I didn't know I have not pushed the changes. I will push in a bit.

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,83 @@
...
+  context "when user votes another user's resource" do
+    before do
+      request.headers['Authorization'] = "Token token=#{user2.refresh_token}"
+    end
+
+    it "upvotes another user's resource" do
+      post :upvote, {resource_name: 'questions', resource_id: 1}
+      expect(response.body).to eq "{\"response\":1}"
+      expect(response.status).to eq 200
+    end
```
andela-oojewale: If you don't mind leveraging on already existing helpers. You could check out the `spec/support` dir. There a number of helpers already provided. E.g `ValidRequest` for valid token and `parsed_json` for response. So, here we may not need to do  `(response.body).to eq "{\"response\":1}".

spec/controllers/votes_controller_spec.rb
```diff
@@ -1,4 +1,83 @@
...
+      expect(response.status).to eq 400
+    end
+
+    it "doesnt downvote another user's resource" do
+      post :downvote, {resource_name: 'answers', resource_id: 1}
+      expect(response.body).to include "Not qualified to vote"
+      expect(response.status).to eq 400
+    end
+  end
+
```
andela-oojewale: I really can't tell where the error lies. My best suggestion would be the controller but this is supposed to return a status code of 403 not 400. Actually, all of these 400 statuses.

andela-echigbo: Yes, you are right @andela-oojewale, and [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) is more details on the HTTP status codes. We have it set in the `ApplicationController#invalid_request`, so @andela-aadepoju you might want to help us fix that in this **PR** . Another consideration would be if it will break something in the front-end.

andela-oakinniranye:
👍

spec/controllers/votes_controller_spec.rb
andela-oakinniranye: i think you should look into @andela-oojewale comment above. There are helpers that would be useful to make an `authenticated` request there. Check the `spec/support/valid_request_helper.rb`

andela-ooranagwa:
👍

PR #85, opened 2016-05-04
Set default index to sort by tags user is subscribed to

andela-oakinniranye:
This PR updates the default index route to fetch all questions ordered by the tags the user is subscribed to.
This eliminates the multiple BE calls the FE has to do in order to fetch questions, when:
1. They have to fetch questions in tags the user is subscribed to
2. They have to also fetch all other questions and append it to the ones they had.
This wouldn't have been so effecient.
Also, configured Sidekiq for test and switched to MockRedis in test mode, so jobs are not queued during tests.
Optimization task: Make `OrderBySubscriptionQuery` to inherit from `UserQuestionsQuery`
Other thoughts: Would it make more sense to still keep supporting the personalized question endpoint?


andela-ooranagwa:
💭 Modifying the question `index endpoint` to return `user custom questions` might make it impossible to return all the questions. My suggestion would be to still allow an endpoint to fetch `all` questions irrespective of `user` subscriptions and have another the returns just the user subscriptions.

app/models/queries/order_by_subscription_query.rb
```diff
@@ -0,0 +1,45 @@
+module Queries
+  class OrderBySubscriptionQuery < UserQuestionsQuery
+    attr_reader :user
+
+    def initialize(user)
+      @user = user
+    end
```
andela-ooranagwa: I think we can also leverage on the **inheritance** and make user of the `initializer` in the `UserQuestionsQuery`

app/models/queries/order_by_subscription_query.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      def sort_by_views
+        "questions.views DESC"
+      end
+
+      def sort_by_date_created
+        "questions.created_at DESC"
+      end
+
+      def case_handler
```
andela-ooranagwa: We might also need this commented out for now also

andela-oojewale:
👍

andela-oakinniranye:
@andela-ooranagwa I think that makes sense, hence the reason the route returns all questions by default, the only thing that's changing here is the ordering of the questions the way it was asked, however, with the route we can still fetch all questions

andela-ooranagwa:
My fear is that it might be impossible for users to view available question untainted by their choice. Unless I'm misunderstanding the `order_by_user_subscription` method, say a user wants to see the state of activity on the app, he/she might have a hard time identifying the most recent question _assuming the recent questions are not in the tag categories he/she is subscribed to_ This is because questions with his/her tags are always prioritised on all the possible routes for fetching questions.
Also consider that this might also reduces the opportunity for users to discover new tags.

PR #86, opened 2016-05-07
Add tags to response payload after temp_token validation

andela-ooranagwa:
The following modifications are carried out to achieve this:
* Completely replacing the tags partial with tags_properties partial
* Write tests for method to retrieve user information with eager loaded tags in Token model
* Create custom method in Token model to retrieve user information and also eager load user tags
* Update the Token controller to only delete token after user information has be retrieved
* Update the user partial to always include the user tag information
* Update the User and Tag controller to render the new tag partial in place of the tag_properties partial
* Update the JSON schema for tags to use the tag properties schema
* Update test that validated the tag_properties schema to use the updated tag schema


app/controllers/tokens_controller.rb
```diff
@@ -1,16 +1,16 @@
-
- class TokensController < ApplicationController
+class TokensController < ApplicationController
   before_action :set_token, only: [:validate]
   skip_before_action :authenticate_user, only: [:validate]
 
   def validate
-    @user = @token.user
+    @user = @token.get_user
+    @token.try(:destroy)
```
andela-oojewale: 👍

app/models/token.rb
```diff
@@ -7,4 +7,8 @@ class Token < ActiveRecord::Base
   def set_temp_token
     self.temp = SecureRandom.uuid.delete('-')
   end
+
+  def get_user
+    User.eager_load(:tags).joins(:tokens).where(id: user_id).first
```
andela-oojewale: 👍

andela-oojewale:
👍 . I think this: ` Completely replacing the tags partial with tags_properties partial` is supposed to be: `Completely replacing the tags_properties partial with tags partial` in the commit message.

app/controllers/tokens_controller.rb
```diff
@@ -1,16 +1,16 @@
...
     @user.update(active: true)
   end
 
   private
     def set_token
-      @token = Token.active.find_by(temp: params[:temp_token]).try(:destroy)
+      @token = Token.active.find_by(temp: params[:temp_token])
```
andela-toduah: Quick question here, is there ever a case where inactive/active tokens share the same `temp` value? If not, we could eliminate having to get active tokens first.

andela-oakinniranye:
This looks good enough, I guess. Would it be possible to slip in this PR a route to retrieve all subscribable tags(`tags without a ParentID`)

andela-ooranagwa:
@andela-oakinniranye I'll see what I can do.

app/controllers/tokens_controller.rb
andela-ooranagwa: Nice catch @andela-toduah. I can't recall any scenario that will require enforcing the `active` constraint. Moreso since **temp_tokens** are deleted immediately, it's arguable that a token will never be `inactive`. I've made the necessary corrections.

PR #87, opened 2016-05-11
Implemented the notification payload for the notification system, with the tests and others

andela-oakinniranye:
After a question, comment or answer is created on Zhishi, we would push the notification to the Notification system to inform all the subscribers of this new resource.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,0 +1,32 @@
+guard :rspec, cmd: "bundle exec rspec" do
+  require "guard/rspec/dsl"
+  dsl = Guard::RSpec::Dsl.new(self)
+
+  # RSpec files
+  rspec = dsl.rspec
+  watch(rspec.spec_helper) { rspec.spec_dir }
+  watch(rspec.spec_support) { rspec.spec_dir }
+  watch(rspec.spec_files)
+
+  # Ruby files
+  ruby = dsl.ruby
+  dsl.watch_spec_files_for(ruby.lib_files)
+
+  # Rails files
+  rails = dsl.rails(view_extensions: %w(jbuilder))
+  dsl.watch_spec_files_for(rails.app_files)
+  dsl.watch_spec_files_for(rails.views)
+
+  watch(rails.controllers) do |m|
+    [
+      rspec.spec.("routing/#{m[1]}_routing"),
+      rspec.spec.("controllers/#{m[1]}_controller"),
+    ]
+  end
+
+  # Rails config changes
+  watch(rails.spec_helper)     { rspec.spec_dir }
+  watch(rails.routes)          { "#{rspec.spec_dir}/routing" }
+  watch(rails.app_controller)  { "#{rspec.spec_dir}/controllers" }
+
+end
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -1,6 +1,7 @@
 class Comment < ActiveRecord::Base
   include VotesCounter
   include ModelJSONHashHelper
+  include NewNotification
 
   has_many :votes, as: :voteable, dependent: :destroy
   belongs_to :comment_on, polymorphic: true, counter_cache: true, touch: true
@@ -12,10 +13,18 @@ class Comment < ActiveRecord::Base
     includes(:user).with_votes
   end
 
+  def on_answer?
+    comment_on_type == 'Answer'
+  end
+
+  def on_question?
+    comment_on_type == 'Question'
+  end
+
   def question
-    if comment_on_type == 'Answer'
+    if on_answer?
       comment_on.question
-    else
+    elsif on_question?
       comment_on
     end
   end
@@ -25,5 +34,17 @@ class Comment < ActiveRecord::Base
       only: [:content],
       include: user_attributes
     )
+  end
+
+  def parents
+    comment_parents = [:question]
+    comment_parents.push(:comment_on) if on_answer?
+    @comment_parents ||= comment_parents.map do |parent|
+      send(parent)
+    end
+  end
+
+  def participants_involved_in_comment
+    parents.map(&:user)
   end
 end
--- a/app/models/concerns/new_notification.rb
+++ b/app/models/concerns/new_notification.rb
@@ -1,0 +1,16 @@
+module NewNotification
+  extend ActiveSupport::Concern
+  included do
+    after_create :notify
+  end
+
+  def notify
+    NotificationSystemWorker.perform_async(model_name.to_s, id)
+  end
+
+  def object_for_notification
+    klass = model_name.to_s
+    serializer = "Notifications::New#{klass}Serializer".constantize
+    serializer.new(self).as_json
+  end
+end
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -1,5 +1,7 @@
 class Tag < ActiveRecord::Base
   include Searchable
+  # NOTE we may need to watch for new tags to alert tags admin to check it out too
+  # include NewNotification
 
   validates :name, presence: true
   has_many :resource_tags
@@ -87,6 +89,10 @@ class Tag < ActiveRecord::Base
 
   def update_parent(representative)
     update(representative: representative)
+  end
+
+  def representative
+    super || self
   end
 
   private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 class User < ActiveRecord::Base
   include AndelaValidator
   include ActionView::Helpers::DateHelper
+  include NewNotification
+
   has_many :comments
   has_many :questions
   has_many :answers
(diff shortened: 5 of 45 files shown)


lib/zi_notification/client.rb
```diff
@@ -0,0 +1,26 @@
...
+        request(:get, path, options)
+      end
+
+      def put(path, options = {})
+        request(:put, path, options)
+      end
+
+      def delete(path, options = {})
+        request(:delete, path, options)
+      end
```
andela-oojewale: I think we can have these verbs as an array and use `define_method`.

app/models/comment.rb
```diff
@@ -26,4 +35,16 @@ def as_indexed_json(_options = {})
       include: user_attributes
     )
   end
+
+  def parents
+    comment_parents = [:question]
```
andela-oojewale: Is this an array because there could be more `parents` added later?

andela-oakinniranye: True @andela-oojewale, the question method is defined on the comment to enable easy access to the question from the comment, wherever it was made, since if it was a comment on a question or answer, they all belong to a question  directly or indirectly.

lib/zi_notification/client.rb
andela-oakinniranye: True and updated.

Guardfile
```diff
@@ -0,0 +1,32 @@
...
+      rspec.spec.("controllers/#{m[1]}_controller"),
+    ]
+  end
+
+  # Rails config changes
+  watch(rails.spec_helper)     { rspec.spec_dir }
+  watch(rails.routes)          { "#{rspec.spec_dir}/routing" }
+  watch(rails.app_controller)  { "#{rspec.spec_dir}/controllers" }
+
+end
```
andela-ooranagwa: Could you throw more light on this. I haven't used `guard` before so I won't want to assume I understand what's happening here (and why) ❓

app/models/comment.rb
```diff
@@ -12,10 +13,18 @@ def self.with_associations
...
+  def on_question?
+    comment_on_type == 'Question'
+  end
+
   def question
-    if comment_on_type == 'Answer'
+    if on_answer?
       comment_on.question
-    else
+    elsif on_question?
```
andela-ooranagwa: 💭

app/models/tag.rb
```diff
@@ -89,6 +91,10 @@ def update_parent(representative)
     update(representative: representative)
   end
 
+  def representative
+    super || self
+  end
+
```
andela-ooranagwa: Hmm... could there be an alternative to this? I think some queries in some other sections of the app fetch reps by looking for tags with `nil` values in their rep column.

app/models/user.rb
```diff
@@ -1,6 +1,8 @@
 class User < ActiveRecord::Base
   include AndelaValidator
   include ActionView::Helpers::DateHelper
+  include NewNotification
+
```
andela-ooranagwa: `NewNotification` sports a very active **after_save**; including it here should cause it to trigger whenever users are saved. I don't think that's desirable or is there other use cases I'm missing?

app/serializers/notifications/user_serializer.rb
```diff
@@ -0,0 +1,14 @@
...
+    private
+      def url
+        user_url(object)
+      end
+
+      def type
+        'User'
+      end
+  end
+end
```
andela-ooranagwa: Nice work with the serializers 👍

lib/zi_notification/client.rb
```diff
@@ -0,0 +1,16 @@
+module ZiNotification
+  # NOTE we still need to implement other interfaces to the different notifications actions
+  module Client
+    class << self
+      [:post, :get, :put, :delete].each do |http_method|
+        define_method http_method do |path, options = {}|
```
andela-ooranagwa: 👯

spec/serializers/notifications/shared_examples/shared_answer_serializer.rb
```diff
@@ -0,0 +1,44 @@
...
+    context "within the answer" do
+      let(:subject) { serialized_answer[answer_root]  }
+
+      it { should have_key(:id) }
+      it { should have_key(:content) }
+      it { should have_key(:user) }
+      it { should have_key(:url) }
+      it { should have_key(:type) }
+      it { should have_key(:question) }
+
```
andela-ooranagwa: JSON schema matchers might be handy 😉

spec/serializers/notifications/shared_examples/shared_question_serializer.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect(subject[:type]).to eq(question_type)
+      end
+
+      context "user properties are defined" do
+
+        it_behaves_like :shared_user_serializer do
+          let(:user) { question.user }
+          let(:user_root) { :user }
+          let(:user_type) { 'User' }
+          let(:serialized_user) { serialized_question[question_root] }
```
andela-ooranagwa: There doesn't seem to be any expectation here.

spec/support/zi_notify_connection_mock.rb
```diff
@@ -0,0 +1,14 @@
...
+    end
+  end
+end
+
+
+RSpec.configure do |config|
+  config.before(:each, mock_notification: true) do
+    allow(ZiNotification::Connection).to receive(:connection).and_return(ZiNotifyConnectionMock)
+  end
+end
```
andela-ooranagwa: 👷

app/models/concerns/new_notification.rb
```diff
@@ -0,0 +1,16 @@
...
+  extend ActiveSupport::Concern
+  included do
+    after_create :notify
+  end
+
+  def notify
+    NotificationSystemWorker.perform_async(model_name.to_s, id)
+  end
+
+  def object_for_notification
```
andela-ooranagwa: Any chance of `eager_loading` all `eager_loadable`? There're some places where we've got nested queries 😨

andela-ooranagwa: *Gracias* for submitting tests alongside 💯

andela-ooranagwa:
👍

Guardfile
andela-oakinniranye: This was the configuration from one other project, but it's basically explicitly telling guard the files to watch for changes and automatically running the tests. Yes, using the default guard configuration would work, but that could also slow down the time taken. Basically guard watches for changes in your filesystem and runs your tests automatically, Also, it's not only useful for running tests, it could be configured to trigger other stuff when changes occur, eg You want to receive an email alert when there is a change in file etc

app/models/tag.rb
andela-oakinniranye: True, I've updated most of the methods fetching the reps to check rather for reps_id. This was just a convenient method to always fetch the representative of the tags, which seemed to be useful when doing some larger collections stuff, e.g I think it would be easier to fetch all collections of tags, by calling tags.map(&:representative).

app/models/user.rb
andela-oakinniranye: This is the way the notification system knows when a new user is created on Zhishi. It seems to also be useful when/if we want to be sending some kind of thank you for joining zhishi mail, but largely it's how the NotificationSystem knows to create or map new users.

andela-oakinniranye: In another thought, I think the hook should be `after_create` rather than `after_save` so that notifications are not triggered everytime user updates or does some other stuff on the question

spec/serializers/notifications/shared_examples/shared_question_serializer.rb
andela-oakinniranye: The expectations are defined in the `shared` serializers, the lets are just adding context to the shared behaviours

app/models/concerns/new_notification.rb
andela-oakinniranye: true, would look into that, we shouldn't be slowing down the db, because of some new content and stuff

Guardfile
andela-ooranagwa: I'd read more on Guard and I came off with the impression that it's most useful if you've got some automated tasks like `mimification`, `preprocessing` etc in your project. This will be different with Zhishi; I wonder if we'd appreciate it knowing that it's running _probably_ redundant tests in the background as we try to come to terms with a new feature. Also given that most editors have autosave enabled, I fear for the cascade of test that might queue.

It will be interesting to have it send everyone notifications whenever boss @andela-oakinniranye makes a change to any file, I bet it'll be every 10mins.

andela-oakinniranye: Lol! So, the way the configuration works is that when a change is made, say in your controller, specs, it would run the test corresponding to that particular file and not your entire suite. I think this is a convenient way rather than having to start running `rspec spec/controller/...`. One more thing, is that like foreman, unless guard is initiated by you, it doesn't automatically start, meaning it doesn't impede your development process if you prefer the other ways.

PR #93, opened 2016-05-20
Implement questions fetching by tags.

andela-oojewale:
Add endpoint to fetch questions by tags.
Add model and request test for the endpoint and the corresponding model method.

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -61,6 +61,10 @@ class QuestionsController < ApplicationController
     @questions = Question.search(params[:q])
   end
 
+  def by_tags
+    @questions = Question.by_tags(params[:tags_id])
+  end
+
   private
 
   def set_question
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -62,7 +62,9 @@ class QuestionsController < ApplicationController
   end
 
   def by_tags
-    @questions = Question.by_tags(params[:tags_id])
+    questions = Question.by_tags(params[:tag_ids]).paginate(page: params[:page])
+    @questions = PaginationPresenter.new(questions)
+    render :index
   end
 
   private
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -62,7 +62,7 @@ class QuestionsController < ApplicationController
   end
 
   def by_tags
-    questions = Question.by_tags(params[:tag_ids]).paginate(page: params[:page])
+    questions = Question.with_basic_association.by_tags(params[:tag_ids]).paginate(page: params[:page])
     @questions = PaginationPresenter.new(questions)
     render :index
   end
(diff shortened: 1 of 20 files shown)


app/controllers/questions_controller.rb
```diff
@@ -55,6 +55,10 @@ def search
     @questions = Question.search(params[:q])
   end
 
+  def by_tags
+    @questions = Question.by_tags(params[:tags_id])
```
andela-echigbo: What do you think about reusing the `question/index` here, Since they are basically returning the same thing? And we will also get the `pagination` attributes for free when we pass this `@questions` through the `PaginationPresenter`. Just like we did in the `top_questions` action. 💭

app/models/question.rb
```diff
@@ -59,6 +59,11 @@ def self.with_answers
     includes(:answers)
   end
 
+  def self.by_tags(tags_id)
+    tags_id = tags_id.map(&:to_i)
+    with_basic_association.joins(:tags).where(tags: {id: tags_id}).distinct
```
andela-echigbo: With the `tags_id.map` above, does it mean we are expecting the client to send the **tags** as an array. I thought it was coming as comma separated string. And maybe you will put me through on why we need the `distinct` when I see you. 🙏

andela-oojewale: Yes, it will be an array of ids of tag ids or representative ids.

app/controllers/questions_controller.rb
andela-oojewale: The reason why we can use question index here is because the questions are neither all questions nor the users questions but questions by a particular set of tags. I think the pagination you mentioned makes sense. Thanks.

andela-oojewale: Cool. I see what you talking about - the index jbuilder view. Makes sense. Thanks.

app/models/token.rb
```diff
@@ -9,6 +9,10 @@ def set_temp_token
   end
 
   def get_user
```
andela-ooranagwa: This methods doesn't seem to be used anymore; maybe we should bid it farewell 👋

app/models/queries/question_filter_query.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def tags_that_have_common_representative
+        representative_table[:id].in(tag_ids)
+      end
+
+      def with_the_representative_tags
+        tag_table[:id].in(tag_table.project(tag_table[:representative_id]).where(tag_table[:id].in(tag_ids)))
+      end
+  end
+end
```
andela-ooranagwa: 🙌

docs/questions_endpoints.md
```diff
@@ -374,3 +375,59 @@ Status: 200
...
+            id: 5
+            title: "What is month one all about?"
+            user_id: 12
+            tags:  [{
+                      matches: [ "operations", "Andela"],
+                   }],
+          }
+    ]
+  }
+```
```
andela-ooranagwa: _Gracias_ for modify the docs

andela-ooranagwa:
_Gracias_ for submitting tests alongside 💯

andela-ooranagwa:
👍

PR #94, opened 2016-05-22
Modify DB seed.

andela-oojewale:
Add necessary validation to some models.
Add more resources to db seed.

### DIFF
--- a/db/helpers/seed_helper.rb
+++ b/db/helpers/seed_helper.rb
@@ -1,0 +1,29 @@
+class SeedHelper
+  def objectify(model)
+    return model.titleize.constantize unless model.include? "_"
+    model.split("_").map(&:capitalize).join.constantize
+  end
+
+  def get_offset(model)
+    @model = objectify(model)
+    rand((@model.first.id)..(@model.last.id))
+  end
+
+  %w(user question answer comment tag vote).each do |model|
+    define_method(:"#{model}") do
+      objectify(model).find_by_id(get_offset(model))
+    end
+  end
+
+  def content
+    Faker::Lorem.sentence
+  end
+
+  def title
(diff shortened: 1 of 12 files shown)


db/helpers/seed_helper.rb
```diff
@@ -0,0 +1,29 @@
+class SeedHelper
+  def objectify(model)
+    return model.titleize.constantize unless model.include? "_"
```
andela-echigbo: 👍  This is not a problem, though, but I believe the next line alone can take care of both conditions.

andela-oojewale: Can you give it a try in rails console and let m know what you get? Could be helpful.

andela-oojewale: Cool. it worked. Thanks.

andela-echigbo: I just tested it out now. It works well. But I could only test with strings that have no `_` as I don't know any of our models that contain that.

andela-oakinniranye: I don't think you'd need the `split&map`, you could use ActiveSupport::Inflector `.classify` or `.camelize`.

spec/factories/comments.rb
```diff
@@ -1,12 +1,9 @@
 FactoryGirl.define do
   factory :comment do
     content { Faker::Hacker.say_something_smart }
+    association :comment_on, factory: :question
```
andela-ooranagwa: 👍

andela-ooranagwa:
👍

db/helpers/seed_helper.rb
andela-oojewale: I already switched to `.camelize` before i saw your comment. Thanks.

PR #95, opened 2016-05-24
fixed aggregation error on postgres

andela-oakinniranye:


andela-oakinniranye:
quick fix for an aggregation problem on Postgres. One immediate consideration: should we change to postgres for development. The benefit of using the current approach is that we can easily switch db engines, however when/if we switch to Postgres, that benefit will be lost

PR #99, opened 2016-05-24
re-Sidekiq Workers

andela-oakinniranye:
SideKiq gets to run before some resources are fully persisted to the database, which often results in a record not found error.
Changed the perform_async method to instead perform in 5 seconds(which is a reasonable time for a record to have been persisted).
Also added the rspec-sidekiq gem so as to enable testing with the sidekiq helpers/matchers.

### DIFF
--- a/app/models/concerns/new_notification.rb
+++ b/app/models/concerns/new_notification.rb
@@ -5,7 +5,7 @@ module NewNotification
   end
 
   def notify
-    NotificationSystemWorker.perform_async(model_name.to_s, id)
+    NotificationSystemWorker.perform_in(5.seconds, model_name.to_s, id)
   end
 
   def object_for_notification
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -12,7 +12,7 @@
 
 common: &default_settings
   # Required license key associated with your New Relic account.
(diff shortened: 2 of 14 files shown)


config/newrelic.yml
```diff
@@ -12,7 +12,7 @@
 
 common: &default_settings
   # Required license key associated with your New Relic account.
-  license_key: 72d898b4665ce655ede38eedf0aea02192d5f3e3
+  license_key: <%= ENV["NEW_RELIC_LICENSE_KEY"] %>
```
andela-oojewale: Lol

andela-oojewale:
👍

app/models/concerns/new_notification.rb
```diff
@@ -5,7 +5,7 @@ module NewNotification
   end
 
   def notify
-    NotificationSystemWorker.perform_async(model_name.to_s, id)
+    NotificationSystemWorker.perform_in(5.seconds, model_name.to_s, id)
```
andela-ooranagwa: Learning time: What's the advantage of `perform_in`?

config/newrelic.yml
andela-ooranagwa: 🙈

andela-ooranagwa:
👍

PR #101, opened 2016-05-25
Tests for Comment Requests

andela-toduah:
Wrote tests to test the requests being sent related to handling comments.
Requests include those for creating a comment, updating a comment, or deleting a comment.

### DIFF
--- a/spec/requests/comments/updating_comment_spec.rb
+++ b/spec/requests/comments/updating_comment_spec.rb
@@ -1,0 +1,47 @@
+require "rails_helper"
+require "support/comment_shared_contexts"
+
+RSpec.describe "Updating a comment", type: :request do
+  include_context "comment resource helpers"
+
+  context "when a different user tries to update" do
+    subject { create :comment }
+
+    it "fails when comment is not owned by user" do
+      put existing_comment_path,
+          {
+            content: Faker::Hacker.say_something_smart,
+            format: :json
+          }, authorization_header
+
+      expect(response.status).to eq 403
+    end
+  end
+
+  context "when same user tries to update" do
+    subject { create :comment, user: valid_user }
+
+    it "fails when comment text is empty" do
+      put existing_comment_path,
+          {
+            content: "",
+            format: :json
+          }, authorization_header
+
+      expect(response.status).to eq 400
+      expect(JSON.parse(response.body)["errors"]).to eq "Comment body can not be empty!"
+    end
+
+    it "updates successfully when all parameters are valid" do
+      new_text = Faker::Hacker.say_something_smart
+      put existing_comment_path,
+          {
+            content: new_text,
+            format: :json
+          }, authorization_header
+
+      expect(response.status).to eq 200
+      expect(JSON.parse(response.body)["content"]).to eq new_text
+    end
+  end
+end
--- a/spec/requests/comments/updating_comment_spec.rb
+++ b/spec/requests/comments/updating_comment_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 require "support/comment_shared_contexts"
 
 RSpec.describe "Updating a comment", type: :request do
-  include_context "comment resource helpers"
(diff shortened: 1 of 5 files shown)


app/controllers/comments_controller.rb
```diff
@@ -42,7 +42,7 @@ def comment_params
   end
 
   def set_resource
-    resource = comment_params[:resource_name].singularize.camelize.constantize.find_by(id: comment_params[:resource_id])
+    resource = comment_params[:resource_name].singularize.camelize.constantize.find_by(id: comment_params[:resource_id]) rescue nil
```
andela-oakinniranye: Do we need the rescue? the `resource_not_found` below is already handling that

andela-oakinniranye:
Great work! Can you look at the `zhishi_url_options` defined on the `[Comment, Answer, Question]` and the `url_for_trackable` defined on the `activity` model to see a way you can use it? rather than duplicating  url params and fix the other comment

app/controllers/comments_controller.rb
andela-toduah: rescue is for situations where the result of `[:resource_name].singularize.camelize` is not a constant. Or if it is, not an ActiveRecord model. Could be caused by an invalid `resource_name` passed in

andela-toduah:
Thanks. Didn't know that was available. I've modified to use `zhishi_url_options`. I don't see how `url_for_trackable` could be used here. Any ideas?

app/controllers/comments_controller.rb
andela-oakinniranye: that would already have been handled by the constraint in the route, I guess!

spec/support/comment_shared_contexts.rb
```diff
@@ -0,0 +1,12 @@
+RSpec.shared_context "comment resource helpers" do
+  let(:resource_name) { subject.comment_on_type.downcase.pluralize }
+  let(:resource_id) { subject.comment_on_id }
+
+  def new_comment_path(options = subject.zhishi_url_options)
+    comments_path(options[:resource_name], options[:resource_id])
```
andela-oakinniranye: I think this would still do the same thing, rather than splitting them up again:
```ruby
    comments_path(subject.zhishi_url_options)
```
Since, it's a hash

spec/requests/comments/updating_comment_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(JSON.parse(response.body)["errors"]).to eq "Comment body can not be empty!"
+    end
+
+    it "updates successfully when all parameters are valid" do
+      new_text = Faker::Hacker.say_something_smart
+      put existing_comment_path,
+          {
+            resource_name: resource_name,
+            resource_id: resource_id,
+            id: subject.id,
```
andela-oakinniranye: here you could do the `zhishi_url_option.merge({content: new_text, format: :json})` and for the others too.
Don't bother about the url_for_trackable, it would be a violation of OOP principles

app/controllers/comments_controller.rb
rahdio: valid point!

andela-toduah: valid point!

spec/requests/comments/updating_comment_spec.rb
andela-toduah: Didn't even need the `resource..` params. That's already gotten from the URL

spec/support/comment_shared_contexts.rb
andela-toduah: valid point!

andela-oakinniranye:
👍

andela-oakinniranye:
One last comment could you change the `"comment resource helpers"` from string to symbol (`:comment_resource_helpers`)? As a mistake that could easily be made would be putting extra spacing between the words or at the boundaries.

PR #105, opened 2016-06-02
fixed circular depenedency upgraded RubyVersion and fixed test errors…

andela-oakinniranye:
… relating to upgrade of Ruby Version


andela-oojewale:
👍

PR #106, opened 2016-06-05
add rake task for remaping tags

andela-aadepoju:
This PR adds a rake task for remapping tags. The task will be carried out by running `rake remap_tags:execute`

### DIFF
--- a/lib/tasks/remap_tags.rake
+++ b/lib/tasks/remap_tags.rake
@@ -1,0 +1,63 @@
+namespace :remap_tags do
+
+  desc "this task remaps and deletes tags. This is a one time task!"
+  task execute: :environment do
+    map_tags
+    rename_and_delete_tags
+    puts "Completed!!!"
+  end
+
+  def map_tags
+    tags = {
+            "food" => ["meals"],
+            "fellows-lagos" => ["fellows lagos", "lagos"],
+            "php" => ["pdo", "laravel", "dbal"],
+            "training" => ["cousera", "simulations"],
+            "sports" =>["football"],
+            "healthcare" => ["kbl", "bupa", "health"],
+            "javascript" => ["dom", "superagent"],
+            "react" => ["react router"],
+            "ruby" => ["rspec", "serializers"],
+            "testing" => ["faker", "mock", "stub", "unit-testing"],
+            "operations" => ["leave", "leave-request", "annual-leave"],
+            "mac" => ["osx"],
+            "nairobi-fellows" => ["brown-bag", "kenya"],
+            "databases" => ["sql", "database-relation"],
+            "api" => ["rest"],
+            "css" => ["flexbox"]
+            }
+
+    tags.each do |key, value|
+      parent_tag = Tag.where(name: key)
+      Tag.where(name: value).update_all(representative_id: parent_tag)
+    end
+    puts "Sucessfuly Mapped Tags!"
+  end
+
+  def rename_and_delete_tags
+    old_tags = {
+                "module" => "ruby",
+                "survey" => "fellows",
+                "staff" => "operations",
+                "queries" => "databases",
+                "deployment" => "laravel"
+                }
+
+    old_tags.each do |key, value|
+        new_tag = Tag.where(name: value)
+        ResourceTag.joins(:tag).where(tags: {name: key}).update_all(tag_id: new_tag)
+    end
+
+    tags_to_delete = ["computer", "operation", "forms", "networking", "pdo-connection",
+     "paid-courses", "programming", "associations", "facility", "ci", "mixin",
+     "discussion", "promotions", "beautiful soup"] + old_tags.keys
+
+    Tag.where(name: "orders").update_all(name: "NY-Office")
+    Tag.where(name: "history").update_all(name: "andela-history")
+    Tag.where(name: "databases").update_all(name: "database")
+
+    Tag.where(name: tags_to_delete).destroy_all
+
+    puts "Sucessfuly Renamed And Deleted Unwanted Tags!!"
+  end
+end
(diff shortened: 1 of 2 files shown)


andela-oojewale:
👍

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+                "queries" => "databases",
+                "deployment" => "laravel"
+                }
+
+    old_tags.each do |key, value|
+        new_tag = Tag.where(name: value)
+        ResourceTag.joins(:tag).where(tags: {name: key}).update_all(tag_id: new_tag)
+    end
+
+    tags_to_delete = ["computer", "operation", "forms", "networking", "pdo-connection",
```
andela-bebowe: I think that should be `computer science` instead of computer

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+            "sports" =>["football"],
+            "healthcare" => ["kbl", "bupa", "health"],
+            "javascript" => ["dom", "superagent"],
+            "react" => ["react router"],
+            "ruby" => ["rspec", "serializers"],
+            "testing" => ["faker", "mock", "stub", "unit-testing"],
+            "operations" => ["leave", "leave-request", "annual-leave"],
+            "mac" => ["osx"],
+            "nairobi-fellows" => ["brown-bag", "kenya"],
+            "databases" => ["sql", "database-relation"],
```
andela-bebowe: its `database relationships` not database-relation

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+    old_tags.each do |key, value|
+        new_tag = Tag.where(name: value)
+        ResourceTag.joins(:tag).where(tags: {name: key}).update_all(tag_id: new_tag)
+    end
+
+    tags_to_delete = ["computer", "operation", "forms", "networking", "pdo-connection",
+     "paid-courses", "programming", "associations", "facility", "ci", "mixin",
+     "discussion", "promotions", "beautiful soup"] + old_tags.keys
+
+    Tag.where(name: "orders").update_all(name: "NY-Office")
```
andela-bebowe: @andela-ooranagwa talked about using small letters all through. shouldn't this apply for NY-Office ?

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+
+    tags_to_delete = ["computer science", "operation", "forms", "networking", "pdo-connection",
+     "paid-courses", "programming", "associations", "facility", "ci", "mixin",
+     "discussion", "promotions", "beautiful soup"] + old_tags.keys
+
+    Tag.where(name: "orders").update_all(name: "NY-Office")
+    Tag.where(name: "history").update_all(name: "andela-history")
+    Tag.where(name: "databases").update_all(name: "database")
+
+    Tag.where(name: tags_to_delete).destroy_all
```
andela-oakinniranye: Are you sure we really want to delete these tags? Basically, the idea of the remapping is that as we increase the number of tags, the system becomes more intelligent to do the mapping for us, which is dependent on us having a repository of existing tags being mapped to each other...

andela-oakinniranye:
Would it be convenient to add an additional task that updates all tags with whitespaces joining them, to use dashes `-` as our conjunctions. e.g `computer science` should be `computer-science`. I think the query should just be:

```ruby
Tag.find_each do |tag|
   name = tag.name.split(' ').join('-')
   tag.update(name: name)
end
```

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+            "sports" =>["football"],
+            "healthcare" => ["kbl", "bupa", "health"],
+            "javascript" => ["dom", "superagent"],
+            "react" => ["react router"],
+            "ruby" => ["rspec", "serializers"],
+            "testing" => ["faker", "mock", "stub", "unit-testing"],
+            "operations" => ["leave", "leave-request", "annual-leave"],
+            "mac" => ["osx"],
+            "nairobi-fellows" => ["brown-bag", "kenya"],
+            "databases" => ["sql", "database-relationships"],
```
andela-bebowe: Before this line we will have to rename `database relationship` to `database-relationships`

andela-bebowe:
@andela-oakinniranye How often do you intend to run that line ? Thinking a validation before creating the tag might be another option

andela-oakinniranye:
Yes, true. It would be ok for this PR to just add the callback as well. However, we still need to run this

andela-bebowe:
So you are saying for this script you will run that and then include the validation in the tag model

andela-oakinniranye:
I meant this PR should add the callback in the Tag model, while this script should also add that as well. Basically this script would only run once as well since we wouldn't always be having these discrepancies in the tag

andela-bebowe:
Ok Nice

lib/tasks/remap_tags.rake
```diff
@@ -0,0 +1,63 @@
...
+    old_tags.each do |key, value|
+        new_tag = Tag.where(name: value)
+        ResourceTag.joins(:tag).where(tags: {name: key}).update_all(tag_id: new_tag)
+    end
+
+    tags_to_delete = ["computer science", "operation", "forms", "networking", "pdo-connection",
+     "paid-courses", "programming", "associations", "facility", "ci", "mixin",
+     "discussion", "promotions", "beautiful soup"] + old_tags.keys
+
+    Tag.where(name: "orders").update_all(name: "NY-Office")
```
andela-bebowe: This NY-Office ought to be ny-office ooh

andela-ooranagwa: There's a `before_save` that will take care of this.

lib/tasks/remap_tags.rake
andela-aadepoju: @andela-ooranagwa what are your thoughts on this?

andela-aadepoju:
I see @andela-ooranagwa has made the changes, thanks boss! 💯

PR #111, opened 2016-06-23
implemented notification queue

andela-oakinniranye:
Implemented the Notification Queue

# How does it work?
When a new resource is posted which affects a user, this implementation would push a notification to each user's queue.

A `GET` request to `/notifications` would return an object with the key `notifications`, which can be looped through to retrieve all notifications for the user. To maintain a manageable system, every queue is limited to 15 elements.

The notification queue is implemented like a plug and play system, such that the library could be used in any other application and the only requirement for a queue owner is to define an `id` method and a resource should define a `queue_tracking_info_json` method.


andela-ooranagwa:
@andela-oakinniranye What do you think of returning only the Notification key and activity url?

andela-oakinniranye:
How would that work? But a reference to the activity would still be required. a typical notification would probably say `your answer: where is the money was upvoted +15` or something like that.

PR #112, opened 2016-07-08
Use andela auth

andela-echigbo:
This Pull Request switches our authentication system to using Andela authentication service.
- The feature is fully tested
- The unused authentication code (controller actions, routes, models) are also removed.
- It also drops the `social_provider` and `token` tables as they are no longer used.


TeatroIO:
I've prepared a stage to preview changes. <a href='http://use-andela-auth.zhishi-project-zhishi-backend-b66fa239ce89d55b2317.ttrcloud.com/' target='_blank'>Open stage</a> or <a href='http://use-andela-auth.zhishi-project-zhishi-backend-b66fa239ce89d55b2317.ttrcloud.com//teatro.log' target='_blank'>view logs</a>.

Gemfile
```diff
@@ -5,9 +5,6 @@ ruby "2.3.1"
 gem 'rails', '4.2.5'
 gem 'rails-api'
 gem 'spring', :group => :development
-gem 'omniauth-oauth2', '~> 1.3.1'
-gem 'omniauth-slack'
-gem "omniauth-google-oauth2"
```
andela-oakinniranye: I thought we planned to have the normal authentication in place, in case Andela Auth isn't an option

andela-echigbo: The idea is that we solely depend on the Andela authentication just like every other Andela services. Do you think there will be a case where this implementation will turn us down in the long run?

app/models/social_provider.rb
```diff
@@ -1,20 +0,0 @@
...
-      sp.uuid= auth.uid
-      sp.profile_picture = auth.info.image
-      url_for_profile = auth.info.urls || {}
-      sp.profile_url = url_for_profile[:Google] || url_for_profile[:Slack] || url_for_profile[:public_profile]
-      sp.email = auth.info.email
-    end
-    social_user.update(user: user) if user && social_user.user.blank?
-    social_user
-  end
-end
```
andela-oakinniranye: Finally getting rid of this old beast 👍 🙇 🙌

spec/requests/shared/shared_authenticated_endpoint.rb
```diff
@@ -1,19 +1,23 @@
 RSpec.shared_examples "authenticated endpoint" do |endpoint, verb, include_answer|
   before(:each) do
+    🙈 = Dummy.create_with_methods({body: "{\"email\": \"#{valid_user.email}\"}"})
+
+    allow_any_instance_of(Faraday::Connection).to receive(:get).and_return(🙈)
```
andela-oakinniranye: oh my lawd! 🙈

app/models/user.rb
```diff
@@ -18,17 +16,13 @@ class User < ActiveRecord::Base
...
-    grabbed = grabbed ? "#{grabbed}%" : email_address
-    user = where("email LIKE :email", email: grabbed).first_or_create do |u|
-      u.name= auth.info.name
-      u.email= auth.info.email
-      u.image = auth.info.image
+  def self.from_andela_auth(user)
+    user = where("email LIKE :email", email: user['email']).first_or_create do |u|
+      u.name= user['name']
+      u.email= user['email']
+      u.image = user['picture']
```
oojewale: This saves the user's image url huh? In the event that someone changes their profile pix, we won't get an update on this field because our query would have found the email and moved on, I guess. Is this what we want?

app/services/cookie_handler.rb
```diff
@@ -0,0 +1,42 @@
+class CookieHandler
+
+  AUTH_URL = 'http://authentication.andela.com/'
+  DATA_PATH = 'loggedin'
+  COOKIE_KEY = 'andela:session'
```
oojewale: I just have a feeling that seeing lines 3 - 5 in the codebase is too much information. I know it shouldn't be hacked but it can also go into `application.yml`.

app/services/cookie_handler.rb
```diff
@@ -0,0 +1,42 @@
...
+    if user.id
+      token = TokenManager.generate_token(user.id)
+      [user, token]
+    end
+  end
+
+  def self.confirm_response(response)
+    begin
+      JSON.parse(response)
+    rescue JSON::ParserError => e
```
oojewale: We're not using the `e`. Guess someone is missing e already. 😂

config/routes.rb
```diff
@@ -37,11 +35,6 @@
         get :tags
         get :activities
       end
-
-      collection do
-        post :logout
-        get :renew_token
-      end
     end
```
oojewale: This is one major benefit of AndelaAuth. Our concerns have been reduced. 💃

db/migrate/20160706122804_drop_social_provider_table.rb
```diff
@@ -0,0 +1,22 @@
+class DropSocialProviderTable < ActiveRecord::Migration
```
oojewale: Sweet

db/migrate/20160706131029_drop_tokens_table.rb
```diff
@@ -0,0 +1,20 @@
+class DropTokensTable < ActiveRecord::Migration
```
oojewale: `sweetable`.

app/services/cookie_handler.rb
andela-oakinniranye: agree to this point, especially line 3, ought to be configured per environment!  authentication-staging.andela.com

app/models/user.rb
andela-oakinniranye: This was a concern raised earlier, not just the picture, but also the name. One way I can immediately think of is to run a check immediately after if there was a change, run an update else return the user

andela-ooranagwa: Nice catch. I'll be modifying it to use `find_or_initialize_by'

PR #116, opened 2017-01-28
#137599493 user authentication

cyrielo:
#### What does this PR do?
This PR solves authentication issues in the backend to work with the new Andela Authentication mechanism
#### Description of Task to be completed?
As a targeted user of Zhishi, I can log into the application in order to GTD.
#### How should this be manually tested?
Running from the localhost:
Open your terminal run `redis-server`
Open a new terminal tab and run the application with this command `rails server` 
Change dir into the frontend folder then run `npm start` to start running the front end
Attempt to login from the login screen after the frontend starts running
#### What are the relevant pivotal tracker stories?
#137599493
#### Screenshots (if appropriate)
N/A
#### Questions:
N/A


app/authenticators/connection/faraday_connection.rb
```diff
@@ -0,0 +1,23 @@
+class Connection::FaradayConnection
+  BASE_ANDELA_URL =  "http://api-staging.andela.com" # use env variable to set the url
```
andela-oakinniranye: please ensure to get urls through an ENV variable

app/controllers/application_controller.rb
```diff
@@ -14,41 +14,30 @@ def invalid_request(message = error_msg, status = 400)
...
-    @token = TokenManager.generate_token(@current_user.id)
-  end
-
-  def logout
-    cookie = request.headers['HTTP_ANDELA_COOKIE']
-    return invalid_request("Cookie must be provided") unless cookie
-    CookieHandler.logout_cookie(cookie)
-    render json: {message: 'logged out'}, status: 200
-  end
+   def login
```
andela-oakinniranye: not sure we'd be needing this route! Also, it creates a source for a bug on the current user.
*What do I mean?*
Imagine you've assigned the `@current_user` in `authenticate_user` and then this is reassigning the value, however at this point, `authenticate_user` might return a `@current_user` or an `unauthorized_token` response. Technically, the `unauthorized_token` should terminate the execution flow, however we want to ensure that only a `User` object is stored as the `current_user`, not just anything! Unless you want to end up checking whether the current_user is a User all over your application. Not sure of why we still need the route though, since auth is being done on the FE!

app/views/application/login.json.jbuilder
```diff
@@ -1,6 +1,5 @@
 json.partial! "users/user", user: @current_user
 json.partial! 'tags/tag', tags: @current_user.tags
```
andela-oakinniranye: ha, I see! this is why you needed the `login`? I believe this was hacked up to serve some other product purposes! We don't really need this anymore, moreover, I believe there should be a route to fetch all the tags of the current user and the current information of the said user

andela-oakinniranye:
Looking good, a few things to fix up! Please run the specs locally, I'm sure the errors are due to some things related to authentication, and since we're decoupling authentication, we need to reimagine the specs. Previous implementation was closely coupled with the auth!

spec/controllers/votes_controller_spec.rb
```diff
@@ -9,64 +9,10 @@
...
-      # expect(response.status).to eq 403
-    end
-  end
-
-  context "when user tries to vote with invalid token" do
-    before do
-      request.headers['Authorization'] = "Token token=badtoken"
-    end
+  context "when user tries to vote with invalid token", invalid_request: true do
+    # before do
```
andela-bmakinwa: @cyrielo remove commented out code.

app/controllers/application_controller.rb
```diff
@@ -14,41 +14,33 @@ def invalid_request(message = error_msg, status = 400)
...
-  def authenticate_user
-    (authenticate_token && authenticate_cookie) || unauthorized_token
-    # authenticate_cookie || unauthorized_token
-  end
-
-  def authenticate_token
-    authenticate_with_http_token do |auth_token, _|
-      user_id = TokenManager.authenticate(auth_token)['user']
-      @token = TokenManager.generate_token(user_id) if user_id
+   def authenticate_user
```
andela-oakinniranye: Not a blocker, but please refactor this method!
I'm thinking you should be able to do something like: 
```ruby
def authenticate_user
  auth = Authenticator.new(request) # pass the request object to the authenticator, you can fetch the details from the request object, i.e let the authenticator determine how it want's to handle it's auth, either by params or header
  # using this approach, you're able to switch authenticators easily! Authenticator only becomes an interface for authentication, that defines 2 public methods: `authenticated?` and `user`.
  # this means when a new version of authentication is to be used, you don't have to touch this controller at all. Also authentication becomes a bit more modular!
  if auth.authenticated?
    # you don't have to define the `create_or_update_user` in the controller
    create_or_update_user(auth.user)
  else
    unauthorized_token
  end
end
```

app/controllers/application_controller.rb
```diff
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::API
 
   attr_reader :current_user
   helper_method :current_user
-  before_action :authenticate_user, except: [:login, :logout]
+  before_action :authenticate_user
 
   def resource_not_found
```
andela-oakinniranye: unless this is a route, move it to the `private` namespace

app/controllers/users_controller.rb
```diff
@@ -2,7 +2,7 @@ class UsersController < ApplicationController
   before_action :set_user, only: [:update, :questions, :tags]
   before_action :set_user_with_activities, only: [:activities]
   before_action :set_user_with_associations_and_statistics, only: [:show]
-  skip_before_action :authenticate_user, only: [:login, :authenticate]
+  skip_before_action :authenticate_user, only: [:authenticate]
```
andela-oakinniranye: not sure you'd be needing this line anymore! since we're trying to abstract authentication

spec/support/andela_auth_mock.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def valid_user
+    @valid_user ||= create(:user)
+  end
+
+  module AuthControllerHelper
+    extend ActiveSupport::Concern
+
+    RSpec.configure do |config|
+      config.include self#, type: :controller
```
andela-oakinniranye: since the callback actions included here are only working for the controller, you should be able to:
`config.include self, type: :controller` and remove the `type: :controller` argument from the callbacks!

PR #117, opened 2017-02-03
#137599493 user authentication

cyrielo:
#### What does this PR do?
This PR creates a new endpoint in users controller to return current user data after authentication
#### Description of Task to be completed?
As a targeted user of Zhishi, I can log into the application in order to GTD.
#### How should this be manually tested?
Running from the localhost:
Open your terminal run `redis-server`
Open a new terminal tab and run the application with this command `rails server` 
Change dir into the frontend folder then run `npm start` to start running the front end
Attempt to login from the login screen after the frontend starts running
#### What are the relevant pivotal tracker stories?
#137599493
#### Screenshots (if appropriate)
N/A
#### Questions:
N/A

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -11,6 +11,11 @@ class UsersController < ApplicationController
   def show
   end
 
+  def me
+    @current_user
+    @token = @current_user.refresh_token
+  end
+
   def questions
     questions = @user.questions.paginate(page: params[:page])
     @questions = PaginationPresenter.new(questions)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -12,8 +12,6 @@ class UsersController < ApplicationController
   end
 
   def me
-    @current_user
-    @token = @current_user.refresh_token
   end
 
   def questions
--- a/app/views/application/login.json.jbuilder
+++ b/app/views/application/login.json.jbuilder
@@ -1,5 +1,0 @@
-json.partial! "users/user", user: @current_user
-json.partial! 'tags/tag', tags: @current_user.tags
-
-# json.user { json.partial! 'users/user', user: @current_user }
-# json.api_key @token
--- a/app/views/users/me.json.jbuilder
+++ b/app/views/users/me.json.jbuilder
@@ -1,4 +1,3 @@
 json.partial! "user", user: @current_user
 json.partial! 'tags/tag', tags: @current_user.tags
-json.extract! @current_user, :email, :active, :created_at, :updated_at,:member_since
-json.set! :token, @token
+json.extract! @current_user, :email, :active, :created_at, :updated_at,:member_since, :token
--- a/app/views/users/me.json.jbuilder
+++ b/app/views/users/me.json.jbuilder
@@ -1,0 +1,4 @@
+json.partial! "user", user: @current_user
+json.partial! 'tags/tag', tags: @current_user.tags
+json.extract! @current_user, :email, :active, :created_at, :updated_at,:member_since
+json.set! :token, @token
--- a/docs/user_endpoints.md
+++ b/docs/user_endpoints.md
@@ -219,10 +219,10 @@ Status: 200
 }
 ```
 
-## GET /users/members/me
+## GET /users/me
 Request
 ```ruby
-  GET /users/members/me
+  GET /users/me
 ```
 
 Response
@@ -232,14 +232,14 @@ Status: 200
   "id": 1,
   "name": "Firstname Lastname",
   "points": 10,
-  "image": "<http://i.imgur.com/GZIxExK.png>",
-  "url": "<http://domain.com/users/1.json>",
+  "image": "http://i.imgur.com/GZIxExK.png",
+  "url": "http://domain.com/users/1.json",
   "tags": ["operations", "Andela"],
-  "email": "<someone@email.com>",
+  "email": "someone@email.com",
   "active": true,
   "created_at": "2017-01-27T13:27:56.997Z",
   "updated_at": "2017-01-27T13:27:57.397Z",
   "member_since": "7 days ago",
-  "token": "<ij2uhhuhj2u.unuodkij.md3o24>"
+  "token": "ij2uhhuhj2u.unuodkij.md3o24"
 }
 ```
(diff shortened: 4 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -12,6 +11,11 @@ def index
   def show
   end
 
+  def me
+    @current_user
```
andela-oakinniranye: One best practice in Rails is to limit to `an` instance variable per action. In this case I'm not sure of the reasons for the redundant call to `@current_user`
Also, you may not need to set the `@token` and `@current_user`!

app/views/application/login.json.jbuilder
```diff
@@ -1,6 +1,5 @@
 json.partial! "users/user", user: @current_user
 json.partial! 'tags/tag', tags: @current_user.tags
```
andela-oakinniranye: please remember we're removing thing file all together

app/views/users/me.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.partial! "user", user: @current_user
+json.partial! 'tags/tag', tags: @current_user.tags
+json.extract! @current_user, :email, :active, :created_at, :updated_at,:member_since
+json.set! :token, @token
```
andela-oakinniranye: a call to `@current_user.token` assigns the token here!

docs/user_endpoints.md
```diff
@@ -217,3 +218,28 @@ Status: 200
...
+```
+
+Response
+```ruby
+Status: 200
+{
+  "id": 1,
+  "name": "Firstname Lastname",
+  "points": 10,
+  "image": "<img_url>",
```
andela-oakinniranye: please in the docs, give an example, shorthand examples like these may be confusing! One should be able to know it's a string, object, array, integer, date!

spec/controllers/users_controller_spec.rb
```diff
@@ -30,6 +30,20 @@
     end
   end
 
+  describe "#me" do
+    let(:user) { valid_user }
+    context "when it is a valid request", valid_request: true do
+      it "gets single user information" do
+        get :me, id: user.id, format: :json
```
andela-oakinniranye: I'd presume this route is: `users/me`, not sure why you need the additional `id` param

andela-oakinniranye: and why not just make a request to the `me` route, since the header already sets the valid user and make your comparisons against `valid_user`

andela-oakinniranye:
Please direct your PR to `develop` branch!

app/views/users/me.json.jbuilder
cyrielo: Do you mean replacing the line to `json.set! :token, @current_user.token`

andela-oakinniranye: yep! In `user.rb` you could alias `refresh_token` to `token` and in the template here, just add `:token` to the `json.extract! @current_user`, i.e.
```ruby
# models/user.rb
def refresh_token
  ....whatever method we had
end
alias :token :refresh_token

# app/views/users/me.json.jbuilder
...
json.extract! @current_user, :email, :active, :token, :created_at, :updated_at,:member_since

```

docs/user_endpoints.md
```diff
@@ -10,6 +10,7 @@ GET /users/:id	| Returns information of a particular user	| False
 GET users/:id/questions| offset, limit ( both could be optional ), user_id, auth_token in header | Returns the questions with the user_id and all the information concerning it  or error message if any.
 GET users/:id/tags| offset, limit ( both could be optional ), user_id, auth_token in header | Returns the tags with the user_id or error message if any.
 GET users/:id/activities| offset, limit ( both could be optional ), user_id, auth_token in header | Returns the all the activities of a user with the user_id or error message if any.
+GET users/member/me| Fetch user info| True
```
andela-oakinniranye: hey, it's not `users/member/me`, it should be `users/me`

spec/controllers/users_controller_spec.rb
cyrielo: the route is currently accessed either via `users/1/me` or `users/member/me`. In a scenario where we don't know the current user id, the latter route will resolve this automatically

docs/user_endpoints.md
cyrielo: Done

spec/controllers/users_controller_spec.rb
andela-oakinniranye: that's not right! `me` route should be on the `collection`, not `member`. in `config/routes.rb`, before your `resources :user`, place `get 'users/me', to: 'users#me'` and remove the current `get :me` on the member.

andela-oakinniranye: you're removing because rails resolves route by precedence, i.e the first declaration that matches would be called

cyrielo: I tried using  ` get 'users/me', to: 'users#me'` too it didn't work at the time I tried, I didn't know about rails resolution by precedence so I had it below the `resources users` this would actually be most preferred method

andela-oakinniranye: Could you try it again? otherwise, it's pointless/hackish going to `get users/1/me`, to fetch the current user, i know it allows you to pass an arbitrary id, but the point is if you're already able to determine the id, why do we need the `/me` route! it's hackish

cyrielo: It now works

docs/user_endpoints.md
```diff
@@ -217,3 +218,28 @@ Status: 200
   ]
 }
 ```
+
+## GET /users/members/me
+Request
+```ruby
+  GET /users/members/me
```
andela-oakinniranye: `GET /users/me`

docs/user_endpoints.md
```diff
@@ -217,3 +218,28 @@ Status: 200
...
+
+Response
+```ruby
+Status: 200
+{
+  "id": 1,
+  "name": "Firstname Lastname",
+  "points": 10,
+  "image": "<http://i.imgur.com/GZIxExK.png>",
+  "url": "<http://domain.com/users/1.json>",
```
andela-oakinniranye: nice, you don't need to include the outermost brackets, it could make someone think he first has to remove them before using

spec/controllers/users_controller_spec.rb
andela-oakinniranye: can you remove the `id: user.id` now?

spec/controllers/users_controller_spec.rb
```diff
@@ -30,6 +30,20 @@
...
+        user = User.first
```
andela-oakinniranye: and you don't need this assignment

andela-oakinniranye:
good cleanup, please fix final comments

andela-oakinniranye:
do you mind adding a request spec?

andela-oakinniranye:
Looks good! 👍

